PR #5, opened 2019-01-17
Add roles to back office users

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-58

Each back office user must have one of a specified set of roles, which determine what they can and cannot do.

The roles are:

- System user
- Administrative super user
- Administrative user
- Data user


irisfaraway:
Currently set to merge into feature/users for the sake of an easy-to-read diff and history – will rebase once that's merged in to master.

PR #6, opened 2019-01-18
Seed users to database

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-58

We need to be able to seed some users so we can test things locally. This PR adds some seed data.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,7 +5,7 @@ WasteExemptionsEngine::Engine.load_seed
 def find_or_create_user(email, role)
   User.find_or_create_by(email: email) do |user|
     user.role = role
-    user.password = ENV["WEX_DEFAULT_PASSWORD"] || "Secret123"
+    user.password = ENV["DEFAULT_PASSWORD"] || "Secret123"
   end
 end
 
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -2,10 +2,21 @@
...
-# The data can then be loaded with the rake db:seed (or created alongside the db with db:setup).
-#
-# Examples:
-#
-#   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
-#   Mayor.create(name: 'Emanuel', city: cities.first)
+def find_or_create_user(email, role)
+  User.find_or_create_by(email: email) do |user|
+    user.role = role
+    user.password = ENV["WEX_DEFAULT_PASSWORD"] || "Secret123"
```
Cruikshanks: I've dropped the convention of prefixing environment variables with the service's acronym for WEX, in the hope that it'll make the code more reusable going forward.

Other than local development (where a .env file would be used anyway) we'll never have WCR and WEX running in the same instance, so there is no need to prefix variables to avoid duplication.

irisfaraway: Makes sense to me! Updated the env vars.

PR #16, opened 2019-01-30
Display regs on dashboard in a useful way

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-59

This PR covers the display and styling of registrations on the dashboard. Each registration should have key information clearly displayed, as well as what actions can be taken.


irisfaraway:
Think this is ready for review, with the qualifications that:

- this may have additional changes to what information is displayed and how after design review
- none of the action links work as there are no actions yet

PR #21, opened 2019-02-04
Activate and deactivate back office user accounts

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-73

System users should have the ability to activate or deactivate all back office user accounts. If a user logs into a deactivated account, they should see a message telling them the account is disabled. They should not be able to access any information or make any changes. Active accounts can access and use the system as normal.

This means we can deal with revoking access to the system without entirely deleting all record of the user (which will be important for version control, etc later on).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -78,7 +78,7 @@ GEM
     defra_ruby_validators (0.1.1)
       activemodel
       rest-client (~> 2.0)
-    devise (4.6.0)
+    devise (4.6.1)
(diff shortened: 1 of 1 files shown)


irisfaraway:
Will be ready for review once https://github.com/DEFRA/waste-exemptions-engine/pull/62 is merged.

irisfaraway:
@Cruikshanks Realised I'd missed an ability check, sorry! Would you mind having a second look?

PR #33, opened 2019-02-12
Invite a new user to the back office

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-70

System users should be able to invite a new user to the WEX back office. When inviting a user, you must provide an email address and a role.

The invited user receives an email asking them to set a password so they can access the system. They must do this to complete the process and finish creating their account.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DEFRA/waste-exemptions-engine
-  revision: 185c48d32d80867570d3afbc9ab4d89dd15ed6e1
+  revision: 0e9f56a81fd30c0a9c5b62b50eada4c23e407557
   branch: master
   specs:
     waste_exemptions_engine (0.0.1)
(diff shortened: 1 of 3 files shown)


spec/requests/user_invitations_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    let(:role) { attributes_for(:user)[:role] }
+    let(:params) do
+      { user: { email: email, role: role } }
+    end
+
+    context "when a system user is signed in" do
+      let(:user) { create(:user, :system) }
+      before(:each) do
+        sign_in(user)
+      end
```
eminnett: Does `sign_in(user)` need to execute before each test or can it be executed once using `before(:context)`? If it is the latter, I suggest making this change for each context in this spec (just to avoid duplicate executions when new test cases are added).

eminnett:
I don't feel I can comment on much of this as I'm not yet familiar with everything this PR touches, but I found a potential inefficiency in the tests.

spec/requests/user_invitations_spec.rb
irisfaraway: @eminnett I think you're right and it's probably overkill. Will make some changes!

irisfaraway: For the record, we tried this out but it resulted in broken tests – will keep this in mind for future specs though.

eminnett: Fine with me.

PR #38, opened 2019-02-14
Make use of airbrake configurable in initializer

Cruikshanks:
Unfortunately the airbrake initializer errors if project_key is not set. The problem is that the initializer is fired in scenarios where we are not actually using the app, for example when running a rake task.

In production when we run rake tasks it's in an environment where environment variables have not been set. As such we need a way to disable using airbrake unless we actually need it.

### DIFF
--- a/config/initializers/airbrake.rb
+++ b/config/initializers/airbrake.rb
@@ -17,8 +17,7 @@
 # In production when we run rake tasks it's in an environment where environment
 # variables have not been  set. As such we need a way to disable using airbrake
 # unless we actually need it.
-use_airbrake = ENV["USE_AIRBRAKE"] == "true" ? true : false
-if use_airbrake
+if ENV["USE_AIRBRAKE"] == "true"
   Airbrake.configure do |c|
     # By default, it is set to airbrake.io As we use our own hosted instance of
     # Errbit we need to set this value


config/initializers/airbrake.rb
```diff
@@ -9,50 +9,61 @@
...
-  # c.blacklist_keys = Rails.application.config.filter_parameters
+
+# Unfortunately the airbrake initializer errors if project_key is not set. The
+# problem is that the initializer is fired in scenarios where we are not
+# actually using the app, for example when running a rake task.
+#
+# In production when we run rake tasks it's in an environment where environment
+# variables have not been  set. As such we need a way to disable using airbrake
+# unless we actually need it.
+use_airbrake = ENV["USE_AIRBRAKE"] == "true" ? true : false
```
eminnett: Why does this need to be a ternary statement? Isn't `ENV["USE_AIRBRAKE"] == "true"` the same as `ENV["USE_AIRBRAKE"] == "true" ? true : false`?

Cruikshanks: Good point that will certainly simplify things.

Though point of order @eminnett ; remember to wait for a review request before commenting on PR's. Until that request goes out, the PR is still considered to be WIP and the dev may not be done 😁 .

eminnett: Ah sorry. I received a notification and thought it had been requested. Sorry about that :-(.

Cruikshanks: I've been caught out by those 'PR Created' notifications as well. I finally learnt my lesson after jumping in too early on a bunch of @irisfaraway PR's 😳

eminnett: I just updated my Trailer settings so hopefully it will only notify when I've been referenced in the PR rather than someone simply creating one ;-).

PR #48, opened 2019-02-18
Remove dummy secret key logic in application.rb

Cruikshanks:
The dummy secret logic was necessary because of the way we were previously deploying environment variables to our environments.

It finally clicked that to support this previous way, we were not only implementing hacks in our app code, we were also maintaining additional customisations and routines in our deployment code.

So a change has been made to how environment variables are deployed meaning these hacks and customisations are no longer needed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,10 @@ gem "govuk_template", "~> 0.23"
 gem "jquery-rails"
 # Use postgresql as the database for Active Record
 gem "pg", "~> 0.18.4"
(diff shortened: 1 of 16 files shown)


Cruikshanks:
@eminnett @irisfaraway This and PR #47 are on a branch that is currently being deployed into our environments.

These changes support the deployment, however if you have issues still feel free to raise them as they can be discussed on my return. The environment isn't going anywhere 😁

PR #56, opened 2019-02-22
Add registration details page

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-60

This must work for both existing registrations, and registrations which are still being created (transient registrations).

The page should display all stored information about the registration, including its exemptions. It should also link to the actions which can be taken for individual exemptions or the entire registration.

Users must be logged in to access the page.


irisfaraway:
These pages should now be in a state where all the information is at least displayed – now we need to work out how to display it in a useful way.

irisfaraway:
Not ready for review but I want to resolve some merge conflicts now rather than later!

PR #62, opened 2019-02-25
Search for in-progress new registrations

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-154

Logged-in users should be able to choose between searching existing registrations or ones which are still in progress (transient registrations). The results should be displayed in a way that is consistent with active registrations, with missing data dealt with gracefully.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DEFRA/waste-exemptions-engine
-  revision: 537848d63f231e9344aef5b28c70feffbfddf59c
+  revision: 650fd955384007cbaa4162c1082f3500cfe74a2b
   branch: master
   specs:
     waste_exemptions_engine (0.0.1)
@@ -73,7 +73,7 @@ GEM
     crass (1.0.4)
     database_cleaner (1.7.0)
     debug_inspector (0.0.3)
-    defra_ruby_style (0.0.2)
+    defra_ruby_style (0.0.3)
       rubocop
     defra_ruby_validators (0.1.1)
       activemodel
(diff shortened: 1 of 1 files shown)


irisfaraway:
I disagree with CodeClimate on this one – think the duplication is necessary in this case.

app/models/waste_exemptions_engine/address.rb
```diff
@@ -4,10 +4,6 @@
 
 module WasteExemptionsEngine
   class Address
-    scope :search_for_postcode, lambda { |term|
-      where("UPPER(postcode) LIKE ?", "%#{term&.upcase}%")
-    }
-
-    scope :site, -> { where(address_type: 3) }
+    include WasteExemptionsEngine::CanBeSearchedLikeAddress
```
Cruikshanks: I don't think it needs the Module name

```suggestion
    include CanBeSearchedLikeAddress
```

I could be wrong 😁

app/models/waste_exemptions_engine/transient_address.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require WasteExemptionsEngine::Engine.root.join("app", "models", "waste_exemptions_engine", "transient_address")
+
+module WasteExemptionsEngine
+  class TransientAddress
+    include WasteExemptionsEngine::CanBeSearchedLikeAddress
```
Cruikshanks: I don't think it needs the Module name

```suggestion
    include CanBeSearchedLikeAddress
```

Again I could be wrong 😀

app/models/waste_exemptions_engine/address.rb
irisfaraway: You are correct! I got it right for the others so not sure why I did it differently here 🤔

Cruikshanks:
I'm also happy if you want to mark the codeclimate issues as `Wontfix` in this case.

PR #69, opened 2019-03-07
Create public register exporter

eminnett:
https://eaflood.atlassian.net/browse/RUBY-205

WEX needs a public register export feature so the public register can be updated via a cron task. The cron task should run every night and push a CSV file to S3.

### DIFF
--- a/config/initializers/defra_ruby_exporters.rb
+++ b/config/initializers/defra_ruby_exporters.rb
@@ -7,14 +7,18 @@ module DefraRuby
     AWS_REGION = (ENV["AWS_REGION"] || "eu-west-1")
 
     EPR_EXPORT_AWS_CREDENTIALS = Aws::Credentials.new(
-      (ENV["AWS_DAILY_EXPORT_ACCESS_KEY_ID"] || raise("Environment variable AWS_DAILY_EXPORT_ACCESS_KEY_ID has not been set")),
-      (ENV["AWS_DAILY_EXPORT_SECRET_ACCESS_KEY"] || raise("Environment variable AWS_DAILY_EXPORT_SECRET_ACCESS_KEY has not been set"))
+      (ENV["AWS_DAILY_EXPORT_ACCESS_KEY_ID"] || raise_missing_env_var("AWS_DAILY_EXPORT_ACCESS_KEY_ID")),
+      (ENV["AWS_DAILY_EXPORT_SECRET_ACCESS_KEY"] || raise_missing_env_var("AWS_DAILY_EXPORT_SECRET_ACCESS_KEY"))
     )
 
-    EPR_EXPORT_S3_BUCKET = (ENV["AWS_DAILY_EXPORT_BUCKET"] || raise("Environment variable AWS_DAILY_EXPORT_BUCKET has not been set"))
+    EPR_EXPORT_S3_BUCKET = (ENV["AWS_DAILY_EXPORT_BUCKET"] || raise_missing_env_var("AWS_DAILY_EXPORT_BUCKET"))
 
     EPR_EXPORT_FILENAME = "waste_exemptions_epr_daily_full.csv"
 
     EPR_EXPORT_TIME = (ENV["EXPORT_SERVICE_EPR_EXPORT_TIME"] || "1:05")
+
+    def self.raise_missing_env_var(variable)
+      raise("Environment variable #{variable} has not been set")
+    end
   end
 end
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # Use this file to easily define all of your cron jobs.
 #
 # It's helpful, but not entirely necessary to understand cron before proceeding.
--- a/config/application.rb
+++ b/config/application.rb
@@ -33,7 +33,7 @@ module WasteExemptionsBackOffice
     # Do not swallow errors in after_commit/after_rollback callbacks.
     config.active_record.raise_in_transactional_callbacks = true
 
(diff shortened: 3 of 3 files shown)


eminnett:
I managed to fix the VCR issue in Travis by writing a custom matcher for VCR. Not great, but about as good as I could do given the trial and error nature of the debugging.

config/schedule.rb
```diff
@@ -0,0 +1,20 @@
...
+set :job_template, "/bin/bash -l -c 'eval \"$(rbenv init -)\" && :job'"
+
+# Only one of the AWS app servers has a role of "db"
+# see https://gitlab.envage.co.uk/dst-south/waste-exemptions-deployment/blob/master/config/deploy.rb#L12
+# so only creating cronjobs on that server, otherwise all jobs would be duplicated everyday!
+
+# This is the daily EPR export job. When run this will create a CSV export of
+# all records and put this into an AWS S3 bucket from which Epimorphics (the
+# company that provides and maintains the EPR) will grab it
+every :day, at: "1:05am", roles: [:db] do
```
Cruikshanks: Is there anyway we can make the time configurable through an ENV VAR? The reason is to confirm it actually works in our non-production environments, as these get switched off everyday at 7pm. So this will never have the chance to run until we ship to production.

lib/defra_ruby/exporters/registration_export_service.rb
```diff
@@ -0,0 +1,74 @@
...
+
+        Rails.logger.info "Finished daily upload of #{file_path} to s3"
+
+        result
+      end
+
+      private
+
+      def full_path
+        Rails.root.join "private", "exports", EPR_EXPORT_FILENAME
```
Cruikshanks: Appreciate this is modelling the behaviour from the previous version of the service. I just don't think we need to create a whole new folder for our exports; `tmp` is sufficient. No strong opinion on whether its `tmp/exports` or just `tmp`, I just don't think we need a new folder for this task.

lib/defra_ruby/exporters/registration_export_service.rb
```diff
@@ -0,0 +1,74 @@
...
+module DefraRuby
+  module Exporters
+    class RegistrationExportService
+
+      EPR_EXPORT_FILENAME = "waste_exemptions_epr_daily_full.csv"
+
+      def epr_export
+        file_path = full_path
+        write_to_file(RegistrationEprExportReport, file_path)
+        save_to_s3(:epr, file_path)
```
Cruikshanks: There is no delete after we have uploaded to AWS S3, but I think it would be prudent to do so rather than leaving the file on our instances.

Could we delete the locally generated file as the final step in the export process?

config/initializers/defra_ruby_exporters.rb
```diff
@@ -0,0 +1,14 @@
+module DefraRuby
+  module Exporters
+    BATCH_SIZE = (ENV["EXPORT_SERVICE_BATCH_SIZE"] || 3000).to_i.freeze
+
+    AWS_REGION = (ENV["AWS_REGION"] || "eu-west-1").freeze
+
+    EPR_EXPORT_AWS_CREDENTIALS = Aws::Credentials.new(
+      (ENV["AWS_DAILY_EXPORT_ACCESS_KEY_ID"] || raise("Environment variable AWS_DAILY_EXPORT_ACCESS_KEY_ID has not been set")),
```
Cruikshanks: Nice 😁 👍 . Probably a pattern we should be using for a few other values which don't have defaults in the app.

config/schedule.rb
```diff
@@ -0,0 +1,20 @@
...
+# It's helpful, but not entirely necessary to understand cron before proceeding.
+# http://en.wikipedia.org/wiki/Cron
+
+# Learn more: http://github.com/javan/whenever
+
+set :output, "/home/rails/waste-exemptions-back-office/shared/log/whenever_cron.log"
+set :job_template, "/bin/bash -l -c 'eval \"$(rbenv init -)\" && :job'"
+
+# Only one of the AWS app servers has a role of "db"
+# see https://gitlab.envage.co.uk/dst-south/waste-exemptions-deployment/blob/master/config/deploy.rb#L12
```
Cruikshanks: Also meant to catch that this line needs updating
```suggestion
# see https://gitlab-dev.aws-int.defra.cloud/open/rails-deployment/blob/master/config/deploy.rb#L69
```

eminnett: 👍

lib/defra_ruby/exporters/registration_export_service.rb
eminnett: Fixed in cab4157.

lib/defra_ruby/exporters/registration_export_service.rb
eminnett: Fixed in 168a3da.

config/schedule.rb
eminnett: Fixed in 611f0df.

config/initializers/defra_ruby_exporters.rb
eminnett: 👍

config/initializers/defra_ruby_exporters.rb
```diff
@@ -0,0 +1,18 @@
+module DefraRuby
```
Cruikshanks: Just spotted this doesn't have a `# frozen_string_literal: true` comment at the top.

- think it should have one
- would this remove the need to use `.freeze`?
- more a question, wonder why rubicon isn't picking this up 🤔

eminnett: The freeze statements are redundant. This is fixed in 9105b48.

PR #85, opened 2019-03-14
Allow back office users to complete an in-progress registration

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-199

If an external user starts a registration but isn't able to complete it, and they call NCCC for help, NCCC staff should be able to pick it up and continue working on it.

The "Resume" links for this feature are already in place - this commit should make them do something useful.

Data agents should not be able to access this feature.


app/views/shared/_resource_details.html.erb
```diff
@@ -122,9 +122,9 @@
           <%= link_to t(".actions.deregister"), "#" %>
         </li>
         <% end %>
-        <% if resource.is_a?(WasteExemptionsEngine::TransientRegistration) %>
+        <% if resource.is_a?(WasteExemptionsEngine::TransientRegistration) && can?(:update, resource) %>
```
Cruikshanks: What are your thoughts on moving this to a helper e.g. `can_resume?(resource)`. Just conscious we use it in a couple of places and this might

- centralise the logic
- make the intent clearer in the view

irisfaraway: Makes sense to me – I was wondering about the same thing but got lazy. Will have a look!

PR #92, opened 2019-03-18
Set default assistance_mode

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-200

A registration which is completed in the back-office should set the `assistance_mode` to `full` by default.

This will take effect once DEFRA/waste-exemptions-engine#98 is merged in.


irisfaraway:
Waiting for https://github.com/DEFRA/waste-exemptions-engine/pull/98 to be merged so the build will work.

PR #98, opened 2019-03-20
Fix error when calling `bundle exec whenever`

eminnett:
When executing `bundle exec whenever`, an `uninitialized constant` error is raised. This is because the feature was developed assuming Rails would autoload the dependencies which isn't the case when executing the `whenever` command. 

This PR resolves the issue by removing the dependency on `DefraRuby::Exporters` from `schedule.rb`.

### DIFF
--- a/spec/schedule_spec.rb
+++ b/spec/schedule_spec.rb
@@ -37,9 +37,12 @@ RSpec.describe "Whenever schedule", vcr: true do
   end
 
   it "allows the `whenever` command to be called without raising an error" do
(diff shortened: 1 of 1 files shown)


spec/schedule_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 require "rails_helper"
 require "whenever/test"
+require "open3"
```
Cruikshanks: I never knew about this. It's awesome! 👍

eminnett: 👍

PR #103, opened 2019-03-22
Add http-2 as a dependency

eminnett:
PR 1995 in the AWS SDK (aws/aws-sdk-ruby/pull/1995) resulted in `Unable to load the http/2 gem.` being raised when we run the application. We could live with this until their fix was released if it weren't for the fact that this error breaks the way we test command line scripts and rake tasks (as stderr is no longer empty).

Adding the `http-2` gem as a dependency fixes the issue at hand.


irisfaraway:
The Hakiri CI has stuck, so rescanned it here and we're OK: https://hakiri.io/projects/872027ab1a1a73/stacks/1048265ef93bbb/builds/b80629820cf16a

PR #105, opened 2019-03-22
Add bulk export functionality to DefraRuby::Exporters

eminnett:
https://eaflood.atlassian.net/browse/RUBY-68

The second half of the functionality for the DefraRuby::Exporters library is to add bulk exports. This is the feature to generate CSV files containing all of the registration exemptions data in the database.

### DIFF
--- a/app/controllers/bulk_exports_controller.rb
+++ b/app/controllers/bulk_exports_controller.rb
@@ -4,39 +4,6 @@ class BulkExportsController < ApplicationController
   def show
     authorize! :read, DefraRuby::Exporters::RegistrationBulkExportReport
 
-    define_links
-    define_exported_at_message
-  end
-
-  private
-
-  def define_exported_at_message
-    export_executed_at = DefraRuby::Exporters::BulkExportFile.first&.created_at
-    msg = export_executed_at.nil? ? t(".not_yet_executed") : t(".exported_at", export_executed_at: export_executed_at)
-    @exported_at_message = msg
-  end
-
-  def define_links
-    @bulk_export_links = DefraRuby::Exporters::BulkExportFile.all.map do |bulk_export_file|
-      construct_link_data(bulk_export_file.file_name)
-    end
-
-    @bulk_export_links.sort_by! { |h| h[:start_date] }.reverse!
-  end
-
-  def construct_link_data(file_name)
-    date_range_description = file_name.split("_").last.sub(".csv", "")
-    date_range = DefraRuby::Exporters::Helpers::DateRange.parse_date_range_description(date_range_description)
-    {
-      start_date: date_range.first,
-      url: DefraRuby::Exporters::RegistrationExportService.presigned_url(:bulk, file_name),
-      text: link_text(date_range)
-    }
-  end
-
-  def link_text(date_range)
-    start_month = date_range.first.strftime("%B %Y")
-    end_month = date_range.last.strftime("%B %Y")
-    start_month == end_month ? start_month : "#{start_month} through #{end_month}"
+    @bulk_exports = BulkExportsPresenter.new
   end
 end
--- a/app/presenters/bulk_exports_presenter.rb
+++ b/app/presenters/bulk_exports_presenter.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+class BulkExportsPresenter
+
+  attr_reader :links, :exported_at_message
+
+  def initialize
+    init_exported_at_message
+    init_links
+  end
+
+  private
+
+  def init_exported_at_message
+    export_executed_at = DefraRuby::Exporters::BulkExportFile.first&.created_at
+    msg = I18n.t(".not_yet_executed")
+    msg = I18n.t(".exported_at", export_executed_at: export_executed_at) if export_executed_at.present?
+    @exported_at_message = msg
+  end
+
+  def init_links
+    @links = DefraRuby::Exporters::BulkExportFile.all.map do |bulk_export_file|
+      construct_link_data(bulk_export_file.file_name)
+    end
+
+    @links.sort_by! { |h| h[:start_date] }.reverse!
+  end
+
+  def construct_link_data(file_name)
+    date_range_description = file_name.split("_").last.sub(".csv", "")
+    date_range = DefraRuby::Exporters::Helpers::DateRange.parse_date_range_description(date_range_description)
+    {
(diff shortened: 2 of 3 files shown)


eminnett:
The force push was to fix a typo in the commit message.

lib/defra_ruby/exporters/registration_export_service.rb
```diff
@@ -39,7 +65,7 @@ def write_to_file(report_class, file_path)
         Rails.logger.info "Finished writing #{file_path} from #{exporter_name}"
       end
 
-      def save_to_s3(export_type, file_path)
+      def self.save_to_s3(export_type, file_path)
         Rails.logger.info "Started daily upload of #{file_path} to s3"
```
Cruikshanks: We just need to tweak the log message now this is being used for both exports, as the bulk ones won't be something that's done daily

```suggestion
        Rails.logger.info "Started upload of #{file_path} to s3"
```

lib/defra_ruby/exporters/registration_export_service.rb
```diff
@@ -50,19 +76,38 @@ def save_to_s3(export_type, file_path)
...
-      def full_path(file_name)
+      private_class_method def self.full_path(file_name)
         Rails.root.join "tmp", file_name
       end
 
-      def export_bucket(export_type)
+      private_class_method def self.export_bucket(export_type)
         aws_config = {
-          epr: { credentials: EPR_EXPORT_AWS_CREDENTIALS, bucket: EPR_EXPORT_S3_BUCKET }
+          epr: { credentials: config.epr_export_aws_credentials, bucket: config.epr_export_s3_bucket },
```
Cruikshanks: I'm of the opinion now there there is more work being done here, and we now have a bonafide config class that the generation of the credentials hash could be moved to it. So the content of this method becomes something like

```ruby
        aws_config = config.aws_config(export_type)
        s3 = Aws::S3::Resource.new(
          region: aws_config[:region],
          credentials: aws_config[:credentials]
        )

        s3.bucket(aws_config[:bucket])
```

spec/defra_ruby/exporters/helpers/date_range_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+    context "when the combination of 'start' and 'end' dates and the 'range_months' results in an ambiguous range" do
+      context "and the the range includes January 1st" do
+        subject(:date_ranges) do
+          start_date = Date.new(2019, 2, 5)
+          end_date = Date.new(2019, 2, 17)
+          described_class.generate_date_ranges(start_date, end_date, two_months)
+        end
+
+        it "uses January 1st as a starting point" do
+          expect(date_ranges.count).to eq(1)
```
Cruikshanks: More for consistency than anything else, in the subsequent higher level contexts the expectation on the size of the `date_ranges` array is brought out into its own test, which includes confirming its of type `Array`.

I'd prefer to do the same within these contexts as well.

spec/defra_ruby/exporters/registration_bulk_export_report_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+  describe "#query" do
+    context "when there are multiple registrations with multiple exemptions" do
+      context "and all are active" do
+        before(:each) do
+          registrations.each do |registration|
+            Helpers::Registrations.activate(registration)
+          end
+        end
+
+        it "returns the correct number of records" do
```
Cruikshanks: The `it` message is right and consistent with the EPR tests. However I think we can make the expected behaviour clearer if we said something like

```ruby
it "returns all records" do
```

.env.example
```diff
@@ -26,11 +26,14 @@ EMAIL_SERVICE_EMAIL=""
 EXPORT_SERVICE_BATCH_SIZE=10
 EXPORT_SERVICE_EPR_EXPORT_TIME="1:05"
 EXPORT_SERVICE_CRON_LOG_OUTPUT_PATH='/home/rails/waste-exemptions-back-office/shared/log/'
+EXPORT_SERVICE_BULK_NUMBER_OF_MONTHS=1
+EXPORT_SERVICE_BULK_EXPORT_FREQUENCY="sunday"
```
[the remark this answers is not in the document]
eminnett: Fixed in 8e75d43.

config/routes.rb
```diff
@@ -25,6 +25,10 @@
   resources :registrations, only: :show, param: :reference
   resources :transient_registrations, only: :show, param: :reference, path: "/transient-registrations"
 
+  # Bulk Exports
```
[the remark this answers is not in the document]
eminnett: Fixed in 2a9b5c0.

lib/defra_ruby/exporters/bulk_export_file.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
eminnett: Fixed in 14bca61.

lib/defra_ruby/exporters/registration_export_service.rb
eminnett: Fixed in 2cd2335.

spec/defra_ruby/exporters/registration_bulk_export_report_spec.rb
eminnett: Fixed in ddd1354.

lib/defra_ruby/exporters/registration_export_service.rb
eminnett: Fixed in 078a31f.

spec/defra_ruby/exporters/helpers/date_range_spec.rb
eminnett: We've discussed why I don't want to change this. I've added comments to the spec in 63c3895 explaining this.

app/controllers/bulk_exports_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def show
+    authorize! :read, DefraRuby::Exporters::RegistrationBulkExportReport
+
+    define_links
+    define_exported_at_message
+  end
+
+  private
+
+  def define_exported_at_message
```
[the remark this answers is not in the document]
eminnett: Fixed in 6e3f402.

PR #109, opened 2019-03-25
Refactor the model concerns namespace 

eminnett:
DEFRA/waste-exemptions-back-office-ta/issues/108

We've agreed to only use the `WasteExemptionsEngine` namespace for code that is defined in the `WasteExemptionsEngine` source code.


eminnett:
I'll wait until DEFRA/waste-exemptions-back-office-ta/pull/90 is approved and merged as it includes a model concern that should be fixed as a part of this PR.

PR #118, opened 2019-03-29
Fix 'load constant' and other 500 errors

eminnett:
Alan's testing of the bulk export feature found a 'load constant' error. Ed did some further testing and found a series of other errors that were hidden by the way paths are auto-loaded when executing the RSpec tests.


eminnett:
I'm not convinced the duplication issue raised by CodeClimate is worth fixing.

irisfaraway:
Yes, I'm happy to ignore CodeClimate on this one.

PR #119, opened 2019-04-01
Add ability to deregister a whole registration

eminnett:
https://eaflood.atlassian.net/browse/RUBY-64

Super Agent and System users in the Back-office users can already deregister individual exemptions. This feature will extend this behaviour to allow these users to deregister all the exemptions for an active registration in one step. The behaviour includes the following:

- deregistration cannot be undone
- users can ‘cease’ or ‘revoke’ an active registration
- users must fill in a reason for deregistering
- only active registrations can be deregistered
- when a registration is deregistered, so are all of the active exemptions belonging to it (with the same status and reason)
- if the last remaining exemption for a registration is deregistered, the whole registration is also deregistered
- only super users and system users can do this

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DEFRA/waste-exemptions-engine
-  revision: 1786e2422747e6d281b13bea41cb00c086ef6ceb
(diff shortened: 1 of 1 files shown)


eminnett:
There is a bit of duplication in how the tests are written, but I wanted to get this feature into review as quickly as I could.

PR #120, opened 2019-04-01
Update to support last email cache feature

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-256

[WEX engine PR 112](https://github.com/DEFRA/waste-exemptions-engine/pull/112) added the ability to see the details of the last email sent as JSON.

However to enable it in the app we need to update the engine to the lastest, and then include the configuration in the engine initializer.

This change makes the necessary changes to the app to support using this feature in our non-production environments.

### DIFF
--- a/config/initializers/waste_exemptions_engine.rb
+++ b/config/initializers/waste_exemptions_engine.rb
@@ -22,5 +22,5 @@ WasteExemptionsEngine.configure do |configuration|
   configuration.use_xvfb_for_wickedpdf = ENV["USE_XVFB_FOR_WICKEDPDF"] || "true"
 
   # Last email cache config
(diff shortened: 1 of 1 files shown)


config/initializers/waste_exemptions_engine.rb
```diff
@@ -20,4 +20,7 @@
 
   # PDF config
   configuration.use_xvfb_for_wickedpdf = ENV["USE_XVFB_FOR_WICKEDPDF"] || "true"
+
+  # Last email cache config
+  configuration.use_xvfb_for_wickedpdf = ENV["USE_LAST_EMAIL_CACHE"] || "false"
```
irisfaraway: Think this should be `configuration.use_last_email_cache`.

Cruikshanks: facepalm

PR #122, opened 2019-04-02
Fix data exports accessibility issues

eminnett:
https://eaflood.atlassian.net/browse/RUBY-68

Alan discovered a few accessibility issues in the data exports view.

- The subheading with the exported on date stamp should use a `p` tag instead of an `h4`.
- The exported on date stamp should be in the format "5:14am on 1 April 2019" so it can be read by screen readers.

This PR fixes these issues.

### DIFF
--- a/app/presenters/bulk_exports_presenter.rb
+++ b/app/presenters/bulk_exports_presenter.rb
@@ -16,12 +16,9 @@ class BulkExportsPresenter
   private
 
   def init_exported_at_message
-    export_executed_at = DefraRuby::Exporters::BulkExportFile.first&.created_at
+    export_executed_at = DefraRuby::Exporters::BulkExportFile.first&.created_at&.to_formatted_s(:time_on_day_month_year)
     msg = I18n.t("bulk_exports.show.not_yet_exported")
-    if export_executed_at.present?
-      export_executed_at = export_executed_at.to_formatted_s(:time_on_day_month_year).gsub("  ", " ")
-      msg = I18n.t("bulk_exports.show.exported_at", export_executed_at: export_executed_at)
-    end
+    msg = I18n.t("bulk_exports.show.exported_at", export_executed_at: export_executed_at) if export_executed_at.present?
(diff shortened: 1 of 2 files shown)


app/presenters/bulk_exports_presenter.rb
```diff
@@ -16,7 +16,7 @@ def initialize
   private
 
   def init_exported_at_message
-    export_executed_at = DefraRuby::Exporters::BulkExportFile.first&.created_at
+    export_executed_at = DefraRuby::Exporters::BulkExportFile.first&.created_at&.strftime("%-d %B %Y")
```
irisfaraway: I know you didn't request a review from me but here I am anyway! The engine actually has a time formats initializer with some date presets, so you can do `.to_formatted_s(:day_month_year)` rather than specifying new ones. That's what we've done in the exemptions list on the registration details page.

Cruikshanks: Just to add to that, hadn't meant for the time element to be dropped! The screen reader was fine with that part, and I think it will help users with assessing how 'fresh' the data is.

However I'd stick with using an initializer, it's just you might need to add one that also includes the timestamp (HH:mm should be sufficient) as well date, month and year.

app/views/bulk_exports/show.html.erb
```diff
@@ -5,9 +5,9 @@
     <h1 class="heading-large">
       <%= t(".heading") %>
     </h1>
-    <h4 class="heading-small">
+    <p class="heading-small">
```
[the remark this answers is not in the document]
eminnett: Fixed in b4c1c87.

app/presenters/bulk_exports_presenter.rb
eminnett: Fixed in c67bf02.

eminnett: Fixed in c67bf02.

PR #135, opened 2019-04-16
Fix issues found with confirmation letter

Cruikshanks:
When the change for the confirmation letter generation went through QA a number of issues were found.

- Date format is different to the original
- Filename is different to the original
- Exemptions information is missing code and shows status rather than expiry date
- translation missing error for sole trader business type

The code was also pretty much taken as is from the previous waste exemptions project but we believe there are some refactorings and changes that can be made to make it more consistent with our current approach and standards.

<details><summary>Example of updated document generated by this new code</summary>
<p>
Example confirmation letter
</p>
</details>

### DIFF
--- a/app/presenters/confirmation_letter_presenter.rb
+++ b/app/presenters/confirmation_letter_presenter.rb
@@ -11,7 +11,7 @@ class ConfirmationLetterPresenter < BasePresenter
   end
 
   def date_of_letter
-    Time.now.in_time_zone("London").to_date.to_formatted_s(:day_month_year)
+    Time.now.to_date.to_formatted_s(:day_month_year)
   end
 
   def submission_date
--- a/spec/presenters/confirmation_letter_presenter_spec.rb
+++ b/spec/presenters/confirmation_letter_presenter_spec.rb
@@ -3,7 +3,7 @@
 require "rails_helper"
 
 RSpec.describe ConfirmationLetterPresenter do
-  let(:today) { Time.new(2019, 4, 2).in_time_zone("London") }
+  let(:today) { Time.new(2019, 4, 2) }
   let(:registration) { create(:registration, :with_active_exemptions) }
   subject { described_class.new(registration) }
 


Cruikshanks:
Other than CodeClimate giving me a **no** even though *I added new tests!* this is now ready for peer review.

Apologies it comes across as a big one, but the changes are all directly tied to the confirmation letter PDF functionality, and if it helps some of those changes are me deleting files (always a nice feeling 😁 ).

Any issues let me know.

Gemfile
```diff
@@ -108,6 +107,9 @@ group :test do
   # Generates a test coverage report on every `bundle exec rspec` call. We use
   # the output to feed CodeClimate's stats and analysis
   gem "simplecov", require: false
+  # A gem providing "time travel" and "time freezing" capabilities, making it
+  # dead simple to test time-dependent code.
+  gem "timecop"
```
cintamani: nooooooo againnnn :D :D shall we go with the https://andycroll.com/ruby/replace-timecop-with-rails-time-helpers-in-rspec/? :D :D I will be happy to then do the housekeeping on the other projects :)

app/presenters/confirmation_letter_presenter.rb
```diff
@@ -15,80 +10,78 @@ def sorted_deregistered_registration_exemptions
     registration_exemptions_with_exemptions.where("state != ?", :active).order_by_state_then_exemption_id
   end
 
-  # Used only if the page contains a single letter.
-  def webpage_title
-    I18n.t("confirmation_letter.single_title", reg_number: registration.reference)
+  def date_of_letter
+    Time.now.in_time_zone("London").to_date.to_formatted_s(:day_month_year)
```
cintamani: I don't think we need this since https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/166

spec/factories/address.rb
```diff
@@ -49,9 +49,17 @@
...
-      sequence :country_iso do |n|
-        "country_iso_#{n}"
-      end
+      uprn { nil }
+      premises { nil }
+      street_address { nil }
+      locality { nil }
+      city { nil }
+      postcode { nil }
+      country_iso { nil }
```
cintamani: 🤔 wondering what's the point to set up those as `nil`?

spec/factories/address.rb
```diff
@@ -49,9 +49,17 @@
...
+    end
+
+    trait :site_uses_address do
+      address_type { 3 }
```
cintamani: NIT 
```suggestion
      address_type { :site }
```
This will work thanks to `enum` and is more readable IMO :)

spec/factories/exemption.rb
```diff
@@ -0,0 +1,15 @@
...
+    category { 0 }
+    url { "https://example.gov.uk/guidance/waste-exemptions-using-waste" }
+    summary { "Use of spam in cooking" }
+    description { "Use of spam in cooking using suitable spam rather than virgin spam or spam which has ceased to be spam - for example by cooking spam with chips." }
+    guidance { "This exemption allows you to use suitable spam rather than virgin spam or spam which has ceased to be spam - for example by cooking spam with chips." }
+
+    sequence :code do |n|
+      "F#{n}"
+    end
+  end
```
cintamani: Would be good if we find a way to load the engine factories in the projects spec. We should research such a solution :D

spec/presenters/confirmation_letter_presenter_spec.rb
```diff
@@ -0,0 +1,153 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ConfirmationLetterPresenter do
+  let(:today) { Time.new(2019, 4, 2).in_time_zone("London") }
```
cintamani: Ditto :D https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/166

spec/presenters/confirmation_letter_presenter_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+    end
+  end
+
+  describe "#contact_full_name" do
+    it "returns the registration's contact first and last name attributes as a single string" do
+      expected_name = "#{registration.contact_first_name} #{registration.contact_last_name}"
+
+      expect(subject.contact_full_name).to eq(expected_name)
+    end
+  end
```
cintamani: Duplicated scenario?

Gemfile
Cruikshanks: I'm gonna stick with timecop. I don't think its a major issue using it, and I'd rather do the conversion as a separate exercise rather than hold this one up.

Using timecop here makes it consistent with the other projects, and then we can plan in the conversation to time-helpers happens across projects in sync.

spec/factories/address.rb
Cruikshanks: Think that was how I found it, but as I'm making other changes this makes perfect sense.

spec/factories/exemption.rb
Cruikshanks: I believe the old WEX code does do this so we can take a look at that.

Word of warning, it does make the dev workflow a little more messy. Say you want to add a trait for a test in the back office, you now need to open the waste-exemptions-engine, make the change, push, review, merge, then update this app to the latest to see it.

But lets have a discussion in our next dev meeting cause I'm sure there is something we can do to avoid the duplication.

spec/presenters/confirmation_letter_presenter_spec.rb
Cruikshanks: 🤦‍♂ I will fix this

spec/factories/address.rb
Cruikshanks: The current address factory populates most of the address fields by default. If I didn't set the attributes to `nil` calls to `build(:address, :site)` would give me an object with the address type set to site, the grid reference and description populated, but also everything else filled in.

However this doesn't reflect what actually happens when you complete the site details in the journey. If you use a grid reference everything in the other address fields is set to nil.

So the expectation on `ConfirmationLetterPresenter.site_address_one_liner()` is that if the site has been set using a grid reference, it should return `""` because none of the address fields will have been set.

What I actually want to do is refactor the factories here to copy what you have done in the [WEX engine](https://github.com/DEFRA/waste-exemptions-engine/blob/master/spec/factories/address.rb) which would solve this problem. But I felt a wholesale refactoring of the address factory was beyond the scope of this PR hence these calls to `nil`.

spec/factories/exemption.rb
cintamani: Yeah 🤔 one for discussion :D

PR #141, opened 2019-04-24
Use NewRegistration instead of TransientRegistration when appropriate

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-62

The engine now allows for multiple types of TransientRegistrations, starting with a NewRegistration.

The back office has generally been assuming that TransientRegistrations are only for the new registration journey. Now that we have the NewRegistration class specifically for that, we need to update the back office to use it.

### DIFF
--- a/spec/requests/page_not_found_spec.rb
+++ b/spec/requests/page_not_found_spec.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Page not found", type: :request do
+  describe "GET /this-page-does-not-exist" do
+    it "redirects to `/errors/404`" do
+      rails_respond_without_detailed_exceptions do
+        get "/this-page-does-not-exist"
+
+        expect(response.code).to eq("404")
+      end
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


irisfaraway:
The Gemfile will need to be changed back to `master` once https://github.com/DEFRA/waste-exemptions-engine/pull/138 is merged in.

irisfaraway:
@Cruikshanks @cintamani I've requested a review for this as well, but it will be voided once https://github.com/DEFRA/waste-exemptions-engine/pull/138 is merged so you may need to re-review again. If you'd rather wait, that is cool 👍

Gemfile
```diff
@@ -62,7 +62,7 @@ gem "paper_trail", "~> 10.2.0"
 # Use the waste exemptions engine for the user journey
 gem "waste_exemptions_engine",
     git: "https://github.com/DEFRA/waste-exemptions-engine",
-    branch: "master"
+    branch: "feature/transient-reg-subclasses"
```
cintamani: I don't think this should be part of the changes right? :)

irisfaraway: Nope, I'll have to rip it out once the engine update is merged!

irisfaraway:
@cintamani @Cruikshanks Now that the first batch of engine changes have been merged in, I've switched the branch back to master here – unfortunately that has voided all the reviews. Would you mind having another quick look?

PR #153, opened 2019-05-01
Configure editing permissions

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-62

Only system users and super agents should be able to edit.

This PR switches on the feature flag for editing, but only allows users with the correct roles to access it.

### DIFF
--- a/spec/requests/engine_spec.rb
+++ b/spec/requests/engine_spec.rb
@@ -4,11 +4,13 @@ require "rails_helper"
 
 RSpec.describe "Waste Exemptions Engine", type: :request do
   describe "/start/new" do
+    let(:request_path) { "/start/new" }
+
     context "when a valid user is signed in" do
       before { sign_in(create(:user, :admin_agent)) }
 
       it "returns a 200 response" do
-        get "/start/new"
+        get request_path
         expect(response).to have_http_status(200)
       end
     end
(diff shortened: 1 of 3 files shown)


irisfaraway:
@Cruikshanks @cintamani This is the last one! 😱

spec/requests/engine_spec.rb
```diff
@@ -4,20 +4,22 @@
 
 RSpec.describe "Waste Exemptions Engine", type: :request do
   describe "/start/new" do
+    let(:request_path) { "/start/new" }
```
cintamani: Can we use a path helper here?

spec/requests/engine_spec.rb
```diff
@@ -35,9 +37,32 @@
...
-        get "/start/new"
+        get request_path
         expect(response).to redirect_to(new_user_session_path)
       end
     end
   end
+
+  describe "edit permissions" do
+    let(:registration) { create(:registration) }
+    let(:request_path) { "/edit/#{registration.reference}" }
```
cintamani: Ditto :)

spec/requests/engine_spec.rb
```diff
@@ -35,9 +37,32 @@
...
+        expect(response).to have_http_status(200)
+      end
+    end
+
+    context "when an admin_agent is signed in" do
+      before { sign_in(create(:user, :admin_agent)) }
+
+      it "redirects to the permissions error page" do
+        get request_path
+        expect(response).to redirect_to("/pages/permission")
```
cintamani: Ditto :)

spec/requests/engine_spec.rb
irisfaraway: We could, but this also makes sure the routes are what we expect them to be. Probably a matter of opinion though! Our request specs generally specify the URL instead of using path helpers so I'd rather keep it this way for consistency 👍

PR #172, opened 2019-05-20
Require defra-ruby-exporters gem before Rails tasks are loaded

cintamani:
This will ensure that the Railtie is called and the gem tasks are included in the Rails tasks


Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DEFRA/defra-ruby-exporters
-  revision: fa2915668042dce9d8173d7e8400e5dcdd25aa5b
+  revision: f2a70f72941760dc2fa26e1d8f07a88a6f705a6a
```
cintamani: Will remove this change once this is approved

Rakefile
```diff
@@ -4,5 +4,6 @@
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
 require File.expand_path("config/application", __dir__)
+require "defra_ruby"
```
cintamani:
```suggestion
require "defra_ruby" 
```

Rakefile
```diff
@@ -4,5 +4,6 @@
...
```
cintamani:
```suggestion
require "defra_ruby"
```

PR #180, opened 2019-05-22
Rename deregistered_on to deregistered_at

Cruikshanks:
This is just about keeping things consistent. Where we are recording a timestamp the pattern has been to use `name_of_field_at` rather than `name_of_field_on`.

For example `created_at`, `updated_at`, and `submitted_at`. This change is just about keeping things consistent.

For reference the rename migration is stored in the [waste-exemptions-engine](https://github.com/defra/waste-exemptions-engine) however all functionality related to deregistering is here hence this change.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,7 +10,7 @@ GIT
 
 GIT
   remote: https://github.com/DEFRA/waste-exemptions-engine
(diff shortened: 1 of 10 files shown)


Cruikshanks:
This cannot be merged until https://github.com/DEFRA/waste-exemptions-engine/pull/190 is merged and the engine version bumped in this repo.

PR #182, opened 2019-05-23
Remove wickedpdf gem and configs

cintamani:
part of: https://eaflood.atlassian.net/browse/RUBY-223

The wicked_pdf gem, initializer config and script are already part of the engine code and dependency of this application.
Requiring and installing the gem twice was causing an error generation in the PDF recorded in the ticket above, for which a PDF was generated with wrong content.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,7 +58,7 @@ gem "paper_trail", "~> 10.2.0"
 # Use the waste exemptions engine for the user journey
 gem "waste_exemptions_engine",
     git: "https://github.com/DEFRA/waste-exemptions-engine",
-    branch: "require-wicked-pdf-correctly"
+    branch: "master"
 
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -52,16 +52,13 @@ gem "kaminari", "~> 1.1"
...
-gem "wicked_pdf", "~> 1.2"
-
 # Used for auditing and version control
 gem "paper_trail", "~> 10.2.0"
 
 # Use the waste exemptions engine for the user journey
 gem "waste_exemptions_engine",
     git: "https://github.com/DEFRA/waste-exemptions-engine",
-    branch: "master"
+    branch: "require-wicked-pdf-correctly"
```
cintamani:
```suggestion
    branch: "master"
```

PR #191, opened 2019-06-01
Exemptions Bulk export rake task 

cintamani:

This code contains the 
 - rake task
 - Services for the bulk export
 - Csv parser for creating the file content
 - Presenter for the CSV entries data

Next steps are to implement the AWS export of the file and records updates. This is currently a WIP new gem functionality (see https://github.com/DEFRA/defra-ruby-aws) 
This code is safe to merge to master and since is already a ton is good to open a PR and get this merged in. 
The commits have some kind of logical order, I have approached the problem in a BDD manner in history, hoping to make it less painful to review. 🙇‍♀


app/presenters/exemption_bulk_report_presenter.rb
```diff
@@ -0,0 +1,136 @@
+# frozen_string_literal: true
+
+class ExemptionBulkReportPresenter < BasePresenter
```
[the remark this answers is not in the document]
cintamani: I don't think this is right given the class is a Presenter, so I am going to ignore this.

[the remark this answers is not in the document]
Cruikshanks: Ahh. I hoped using the `BasePresenter` you'd avoid having to have a method per property, but I can see you are naming the methods after the columns hence the need for so many.

So makes sense to ignore codeclimate on this one

app/presenters/exemption_bulk_report_presenter.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  def exemption_deregister_comment
+    deregistration_message
+  end
+
+  def assistance_type
+    registration.assistance_mode
+  end
+
+  def registration_detail_url
```
Cruikshanks: We have an env var for the domain which we then use in `application.rb`, plus the full URL needs to match the new service not the old. For example in our current test environment that would be `https://admin-wex-tst.aws.defra.cloud/registrations/WEX000216`.

See https://github.com/DEFRA/waste-exemptions-back-office-ta/blob/master/config/application.rb#L46 for details of where we bring in the env var. So this in the interim could be replaced with

```ruby
File.join(Rails.configuration.back_office_url, "registrations", registration.reference)
```

app/services/reports/bulk_export_service.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Reports
+  class BulkExportService < ::WasteExemptionsEngine::BaseService
+    def run(*)
+      first_day_of_the_month = starts_from
+
+      while first_day_of_the_month < Date.today
```
Cruikshanks: Like my previous comment, it would be great if we could ensure that any exceptions thrown during the generation of any one file does not cause the service to stop generating the others.

It comes to mind here, but there may be better places to do this.

app/services/reports/monthly_bulk_report_service.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    private
+
+    def temp_file
+      @_temp_file ||= Tempfile.new(file_name)
+    end
+
+    def file_name
+      data_from_date = @first_day_of_the_month.strftime("%Y%m%d")
```
Cruikshanks: Would prefer to stick with the convention of placing all date formatters in `config/initializers/time_formats.rb` and then referring to it here using `to_formatted_s(:my_format)` rather than using `strftime()` directly.

app/models/monthly_bulk_serializer.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  private
+
+  attr_reader :first_day_of_the_month
+
+  def exemptions_data
+    registration_exemptions_scope.find_in_batches(batch_size: batch_size) do |batch|
+      batch.each do |registration_exemption|
+        yield parse_registration_exemption(ExemptionBulkReportPresenter.new(registration_exemption))
```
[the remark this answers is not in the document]
cintamani: Makes sense :)

app/presenters/exemption_bulk_report_presenter.rb
```diff
@@ -0,0 +1,136 @@
...
+  def exemption_status
+    state
+  end
+
+  def exemption_valid_from_date
+    registration_date
+  end
+
+  def exemption_expiry_date
+    # TODO: Implement into the Registration model
```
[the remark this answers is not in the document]
cintamani: Oh wow! Completely missed that! Thanks :)

app/presenters/exemption_bulk_report_presenter.rb
cintamani: Yes, I have left it as a "TODO" as I wasn't sure where to look at :) Thanks for pointing at it! Saved me some time :D

app/services/reports/monthly_bulk_report_service.rb
cintamani: Yes, I have done that for the dates in the presenter but this one didn't look "important" enough? I mean, is not a date format we will reuse elsewhere any time soon. Perhaps I can give it a name that includes the `reports` key :)

cintamani:
@Cruikshanks the reason behind the fact that I didn't scope the models and presenters under `Reports` is because those are objects that can be reused outside of the scope of a Report generation. For example, they can be used to write an interactive report view in the back office one day, so that we don't rely on a background job running once a week. 
Still, the `Reports` scope might make sense anyway and hence I will give it a go :) 

Thanks a lot for your review :)

app/models/reports.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+require_relative "reports/monthly_bulk_serializer"
```
Cruikshanks: Already had a chat about this. Happy to accept we don't know why this is needed in models, but not in presenters or serializers. But could we add a comment owning up to this, something like

```ruby
# We have put everything to do with reports into a `/reports` folder in the
# model, presenters, and services folder mainly just to keep things tidy.
# For presenters and services that was fine, but things broke in models.
#
# Hence this file, which magically seems to fix things! We don't understand why,
# and should we ever get 5 minutes and the inclination to do so we might try
# figure it out. But for now this we're just moving on 😁
```

cintamani:
```suggestion
# We have put everything to do with reports into a `/reports` folder in the
# model, presenters, and services folder mainly just to keep things tidy.
# For presenters and services that was fine, but things broke in models.
#
# Hence this file, which magically seems to fix things! We don't understand why,
# and should we ever get 5 minutes and the inclination to do so we might try
# figure it out. But for now, we're just moving on 😁
require_relative "reports/monthly_bulk_serializer"
```

app/models/reports.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# We have put everything to do with reports into a `/reports` folder in the
+# model, presenters, and services folder mainly just to keep things tidy.
+# For presenters and services that was fine, but things broke in models.
+#
+# Hence this file, which magically seems to fix things! We don't understand why,
+# and should we ever get 5 minutes and the inclination to do so we might try
+# figure it out. But for now, we're just moving on 😁
```
cintamani:
```suggestion
# figure it out. But for now, we're just moving on :D
```

PR #198, opened 2019-06-06
Implement Aws bucket load of report bulk CSV files

cintamani:
Part of: https://eaflood.atlassian.net/browse/RUBY-392

Implements loading the Bulk report files to Aws bucket using the `defra-ruby-aws` gem.

This also adds end-to-end test coverage. Next step is to implement the records update. 

This is also dependent from: https://github.com/DEFRA/defra-ruby-exporters/pull/9 and https://github.com/DEFRA/defra-ruby-aws/pull/4

### DIFF
--- a/app/services/reports/monthly_bulk_report_service.rb
+++ b/app/services/reports/monthly_bulk_report_service.rb
@@ -56,7 +56,7 @@ module Reports
     end
 
     def bucket_name
-      ENV["AWS_BULK_EXPORT_BUCKET"]
+      WasteExemptionsBackOffice::Application.config.bulk_reports_bucket_name
     end
   end
 end
--- a/app/services/reports/monthly_bulk_report_service.rb
+++ b/app/services/reports/monthly_bulk_report_service.rb
@@ -5,27 +5,24 @@ module Reports
     def run(first_day_of_the_month)
       @first_day_of_the_month = first_day_of_the_month
 
-      # Generate report
-      temp_file.write(bulk_report)
+      populate_temp_file
 
-      # upload file to s3
+      load_file_to_aws_bucket
 
-      # rubocop:disable Layout/CommentIndentation
-      # if upload successfull
-        # record content created
-      # else
-        # retry 3 times then exit with failure
-      # end
-      # rubocop:enable Layout/CommentIndentation
+      # record_content_created
     rescue StandardError => e
       Airbrake.notify e, file_name: file_name
-      Rails.logger.error "Generate bulk export csv error for #{file_name}:\n#{error}"
+      Rails.logger.error "Generate bulk export csv error for #{file_name}:\n#{e}"
     ensure
       temp_file.close
       temp_file.unlink
     end
 
     private
+
+    def populate_temp_file
+      temp_file.write(bulk_report)
+    end
 
     def temp_file
       @_temp_file ||= Tempfile.new(file_name)
@@ -40,6 +37,26 @@ module Reports
 
     def bulk_report
       MonthlyBulkSerializer.new(@first_day_of_the_month).to_csv
+    end
+
+    def load_file_to_aws_bucket
+      result = nil
+
+      3.times do
+        result = bucket.load(temp_file)
+
+        break if result.successful?
+      end
+
+      raise(result.error) unless result.successful?
+    end
+
+    def bucket
+      @_bucket ||= DefraRuby::Aws.get_bucket(bucket_name)
+    end
+
+    def bucket_name
+      ENV["AWS_BULK_EXPORT_BUCKET"]
     end
   end
 end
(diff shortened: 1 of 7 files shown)


spec/services/reports/monthly_bulk_report_service_spec.rb
```diff
@@ -5,14 +5,45 @@
...
-        expect(MonthlyBulkSerializer).to receive(:new).and_return(monthly_bulk_serializer)
-        expect(monthly_bulk_serializer).to receive(:to_csv)
+          expect_no_error
 
-        MonthlyBulkReportService.run(Date.today)
+          MonthlyBulkReportService.run(first_day_of_the_month)
+        end
       end
+
+      context "when the request succeed" do
```
Cruikshanks: Think you have a copy+paste fail here. Guessing that should be `context "when the request fails"`

spec/services/reports/monthly_bulk_report_service_spec.rb
```diff
@@ -5,14 +5,45 @@
...
+    def stub_failing_request
+      stub_request(
+        :put,
+        %r{https:\/\/.*\.s3\.eu-west-1\.amazonaws\.com\/20190601-20190630\.csv.+}
+      ).to_return(
+        status: 403
+      )
+    end
+
+    def expect_no_error
```
Cruikshanks: Are you expecting to need this again in subsequent tests? Just querying if it needs to be in a separate method just yet. There seems to be only one call to it, and when we need the negative version we just put the `expect` straight into the test (line 25) rather than also use a method.

app/services/reports/monthly_bulk_report_service.rb
```diff
@@ -41,5 +38,25 @@ def file_name
...
+
+      raise(result.error) unless result.successful?
+    end
+
+    def bucket
+      @_bucket ||= DefraRuby::Aws.get_bucket(bucket_name)
+    end
+
+    def bucket_name
+      ENV["AWS_BULK_EXPORT_BUCKET"]
```
Cruikshanks: We have another one of those unwritten conventions 😳 

Excluding initializers, for any config the app needs we tend to load it once in `config/application.rb` and then refer to `Rails.configuration` within the rest of the app.

I'll be honest this just always seems to have been the case, so I can't say why 😕 but as we seem to have a convention we should probably stick to it for now.

spec/services/reports/monthly_bulk_report_service_spec.rb
cintamani: Ops!! 🙈

spec/services/reports/monthly_bulk_report_service_spec.rb
cintamani: I went for a separate method to avoid adding a comment to the code. Is confusing to see that expectation in line with the test without knowing why we check it. I basically went for the method for documentation purposes.

cintamani: Will move it back in the test and add a comment if you prefer that 🙇‍♀

app/services/reports/monthly_bulk_report_service.rb
cintamani: Will do :)

config/initializers/defra_ruby_aws.rb
```diff
@@ -0,0 +1,22 @@
...
+require "defra_ruby/aws"
+
+DefraRuby::Aws.configure do |c|
+  def raise_missing_env_var(variable)
+    raise("Environment variable #{variable} has not been set")
+  end
+
+  ENV["AWS_BULK_EXPORT_BUCKET"].present? || raise_missing_env_var("AWS_BULK_EXPORT_BUCKET")
+  ENV["AWS_BULK_EXPORT_ACCESS_KEY_ID"].present? || raise_missing_env_var("AWS_BULK_EXPORT_ACCESS_KEY_ID")
+  ENV["AWS_BULK_EXPORT_SECRET_ACCESS_KEY"] || raise_missing_env_var("AWS_BULK_EXPORT_SECRET_ACCESS_KEY")
```
cintamani: Going to remove those as the validation is already performed in the `DefraRuby::Aws::Bucket`

spec/services/reports/monthly_bulk_report_service_spec.rb
Cruikshanks: Its more in the success test we call this method which is actually just the line

```ruby
expect(Airbrake).to_not receive(:notify)
```

However in the failing test rather than call a method there is just the line

```ruby
expect(Airbrake).to receive(:notify).once
```

So the documentation argument makes sense, but then I guess I would expect to see the failing test to be updated as

```ruby
it "fails gracefully and report the error" do
  ...
  expect_an_error
  ...
end

def expect_no_error
  expect(Airbrake).to_not receive(:notify)
end

def expect_an_error
  expect(Airbrake).to receive(:notify).once
end
```

Does this make sense? 😕

cintamani: yep it was inconsistent :/ I went for comments, that's more in line with the rest we have and doesn't make a huge difference

PR #199, opened 2019-06-07
Save report generation to the database

cintamani:
A new model has been created that point to the same existing table as per the exporters gem.
An issue will be created to deal with the table renaming in separate PRs.


app/models/generated_report.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require "defra_ruby/exporters"
+
+class GeneratedReport < ActiveRecord::Base
```
Cruikshanks: Will it be a pain to move this to the `reports` namespace? 😁

cintamani: Oh yeah! 👍

PR #201, opened 2019-06-07
Monthly Bulk Export - Ensure the database entries are cleared

cintamani:

Ensures that the databse is cleared every time a new bunch of file is created.

### DIFF
--- a/app/services/reports/bulk_export_service.rb
+++ b/app/services/reports/bulk_export_service.rb
@@ -3,7 +3,7 @@
 module Reports
   class BulkExportService < ::WasteExemptionsEngine::BaseService
     def run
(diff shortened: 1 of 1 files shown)


app/services/reports/bulk_export_service.rb
```diff
@@ -2,7 +2,11 @@
 
 module Reports
   class BulkExportService < ::WasteExemptionsEngine::BaseService
-    def run(*)
+    def run
+      return if WasteExemptionsEngine::Registration.count == 0
```
cintamani:
```suggestion
      return if WasteExemptionsEngine::Registration.count.zero?
```

PR #203, opened 2019-06-10
Add links to AWS bulk export files

cintamani:

Add links to AWS bulk export files using the new defra_ruby_aws code

[View it](https://user-images.githubusercontent.com/1385397/59194176-12dc1a00-8b80-11e9-9717-3859d085335f.gif)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -83,7 +83,7 @@ GEM
     arel (6.0.4)
     ast (2.4.0)
     aws-eventstream (1.0.3)
-    aws-partitions (1.172.0)
+    aws-partitions (1.173.0)
     aws-sdk-core (3.54.2)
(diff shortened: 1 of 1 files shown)


app/models/reports/generated_report.rb
```diff
@@ -5,5 +5,9 @@
 module Reports
   class GeneratedReport < ActiveRecord::Base
     self.table_name = :defra_ruby_exporters_bulk_export_files
+
+    def starts_from
+      Date.parse(file_name.split("-").first, :plain_year_month_day)
```
cintamani: This is a TODO fix (this introduces a constraint in the file name format that needs reinforcement). Is good to go for an initial dev env test though and the refactoring is almost ready so I would like to merge this in the meantime

PR #211, opened 2019-06-11
Fix namespace for EPR export in schedule.rb

Cruikshanks:

The reference to the Defra exporters EPR export in `schedule.rb` was using the old namespace and hence wouldn't work. This change fixes the reference to use the correct namespace.


Cruikshanks:
Missed updating the tests 🤦‍♂

cintamani:
I have added the fix here with the new rake task: https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/222 :)

PR #213, opened 2019-06-11
Return blank site address when location is by grid reference

cintamani:

Return a nil value for the site address export details when the site is located by a grid reference.


Cruikshanks:
Merging in @cintamani absence

PR #229, opened 2019-06-20
Build first reminder email

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-252

This is the first reminder email that digital users will get. It should be sent 4 weeks before the registration is due to expire.

This story covers building the mailer and putting the HTML and text into the code.

email-screenshot

You can test this with the last registration in the database by running `bundle exec rake reminder:first`. (We'll rip this out once we have something actually triggering it, but for now this is there for testing.)

### DIFF
--- a/app/mailers/renewal_reminder_mailer.rb
+++ b/app/mailers/renewal_reminder_mailer.rb
@@ -38,7 +38,7 @@ class RenewalReminderMailer < ActionMailer::Base
     "TODO"
   end
 
-  def exemptions(_registration)
-    ["TODO"]
+  def exemptions(registration)
+    registration.exemptions.map { |ex| "#{ex.code} #{ex.summary}" }
   end
 end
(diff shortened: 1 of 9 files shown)


app/mailers/renewal_reminder_mailer.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def reference(registration)
+    registration.reference
+  end
+
+  def site_location(registration)
+    address = registration.site_address
+
+    if address.postcode.present?
```
cintamani: Can we use the `Address#located_by_grid_reference?` method here instead? :)

irisfaraway: Ooh, I missed that. Yes, I think that'll work. Will fix!

irisfaraway: Resolved by https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/229/commits/ce9ea69c101cdce5716e37bcfd13a45072fea1bb

irisfaraway:
@cintamani Would you mind having a second look? Fixed the method but it had some minor knock-on affects on an exporter test.

PR #245, opened 2019-06-28
Schedule first email reminder job

cintamani:
From https://eaflood.atlassian.net/browse/RUBY-423

Upldate the schedule config to run the first email reminder job once a day in he morning.

Blocked by: https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/243


config/schedule.rb
```diff
@@ -29,3 +29,10 @@
 every bulk_frequency, at: bulk_time, roles: [:db] do
   rake "reports:generate:bulk"
 end
+
+# This is the daily first renewal reminder mail service.
+# Will run once a day in the early morning hours and send email reminders about
+# registrations that will expire in X time.
+every :day, at: (ENV["FIRST_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME"] || "1:05"), roles: [:db] do
```
irisfaraway: Just some thoughts about naming – are we expecting that the service will be responsible for sending all the reminders, or just the first reminder? I was thinking both, but in that case, this env var name wouldn't be accurate. But if we're having two separate services then I feel like the service should be named more specifically.

spec/schedule_spec.rb
```diff
@@ -42,6 +42,13 @@
     expect(schedule.sets[:output]).to eq(expected_output_file)
   end
 
+  it "takes the EPR execution time from the appropriate ENV variable" do
```
irisfaraway:
```suggestion
  it "takes the email reminder execution time from the appropriate ENV variable" do
```

config/schedule.rb
cintamani: This variable, in particular, can definitely be reused if we want both jobs to run at the same time, but this is not a decision I want to take now. I think we should refactor it later if we decide to reuse it. In regard to the service object, I do believe we will need 2 separate services for the 2 email reminders. The mailer method will differ and so will differ the time when those will be sent, and sending these pieces of information as parameters to the service feels to me like defeating the purpose of the object :D So I think we should rename the Service to make it clearer it is a `First` email reminder service.

config/schedule.rb
```diff
@@ -28,3 +28,10 @@
 every :day, at: bulk_time, roles: [:db] do
   rake "reports:generate:bulk"
 end
+
+# This is the daily first renewal reminder mail service.
+# Will run once a day in the early morning hours and send email reminders about
+# registrations that will expire in X time.
+every :day, at: (ENV["FIRST_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME"] || "1:05"), roles: [:db] do
+  rake "email:renew_reminder:send"
```
cintamani: I will rename this to "email:renew_reminder:first:send" in a separate PR

PR #246, opened 2019-06-28
Rename email reminders task to include `first`

cintamani:
This renames the Email reminders to include `first` so that we can distinguish between the first and second email reminders later on.

Open to suggestion. I am not super super happy about this naming 🤔


cintamani:
Thanks @Cruikshanks :D Me and Iris discussed this already https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/245#discussion_r298531594 :) we agree with the 2 separate services approach but I have left this change out of the other main PRs :)

PR #248, opened 2019-06-28
Fix error when in confirmation letter for expired

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-453

Prior to this change if the registration contained an exemption with a state of `expired` it would error when the user attempted to view the confirmation letter.

A quick check of the code identified that the logic in the `ConfirmationLetterPresenter.registration_exemption_status()` method would attempt to generate a display version of the deregistered date for any exemption without a status of `active`.

The problem is that an expired exemption does not have a deregistered date hence the code was failing. So a quick tweak of the logic and 🎉 we have a working confirmation letter.

### DIFF
--- a/app/presenters/confirmation_letter_presenter.rb
+++ b/app/presenters/confirmation_letter_presenter.rb
@@ -61,7 +61,7 @@ class ConfirmationLetterPresenter < BasePresenter
   end
 
   def registration_exemption_status(registration_exemption)
-    display_date = if registration_exemption.state == "active"
+    display_date = if registration_exemption.state == "active" || registration_exemption.state == "expired"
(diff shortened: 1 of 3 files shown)


app/presenters/confirmation_letter_presenter.rb
```diff
@@ -61,7 +61,7 @@ def exemption_description(exemption)
   end
 
   def registration_exemption_status(registration_exemption)
-    display_date = if registration_exemption.state == "active"
+    display_date = if registration_exemption.state == "active" || registration_exemption.state == "expired"
```
cintamani:
```suggestion
    display_date = if registration_exemption.active? || registration_exemption.expired?
```

PR #252, opened 2019-07-01
Do not seed exemptions in the test suite, use factories instead

cintamani:
This removes the need of seeding Exemption in the database and fixes connected specs.
Housekeeping and alignment with other services.

### DIFF
--- a/spec/models/reports/monthly_bulk_serializer_spec.rb
+++ b/spec/models/reports/monthly_bulk_serializer_spec.rb
@@ -10,9 +10,14 @@ module Reports
 
     describe "#to_csv" do
       it "return a string in CSV format with complete exemptions details for the given month" do
-        create_list(:registration, 3)
-        # The registration factories default to creating 3 exemptions per-registration factory.
+        3.times do
+          registration_exemptions = build_list(:registration_exemption, 3)
+          create(:registration, registration_exemptions: registration_exemptions)
+        end
+
+        # Since every created registration has 3 registration exemptions, then:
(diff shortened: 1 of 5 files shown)


spec/models/reports/monthly_bulk_serializer_spec.rb
```diff
@@ -11,14 +11,12 @@ module Reports
     describe "#to_csv" do
       it "return a string in CSV format with complete exemptions details for the given month" do
         create_list(:registration, 3)
-
-        # TODO: Solve this at a factory level and deal with the cascade of fixes
-        WasteExemptionsEngine::RegistrationExemption.update_all(registered_on: Date.today)
-
+        # The registration factories default to creating 3 exemptions per-registration factory.
+        total_exemptions = 3 * 3
```
irisfaraway: I'm not sure about doing this over `WasteExemptionsEngine::RegistrationExemption.count` - it feels like we are making the test more fragile as if the factory changes how many exemptions it has, then it'll break.

cintamani: Yeah. I think if we change line 13 (https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/252/files#diff-f71d8ac0837fc977208704135536563dR13) to clearly add a number of registration exemptions instead that would make this clearer. WDYT?

irisfaraway: @cintamani Sounds good to me :)

PR #254, opened 2019-07-02
Add feature toggle check for first email reminder

cintamani:
Since we want to include the code in the release but we want to control when the feature is actually active, this uses the new FeatureToggle class in the engine to determine if we should send emails or not. 

Depends on https://github.com/DEFRA/waste-exemptions-engine/pull/222


Cruikshanks:
Holding off reviewing until https://github.com/DEFRA/waste-exemptions-engine/pull/222 is merged in

PR #255, opened 2019-07-02
Add rake task to set the status of expired registration exemptions

cintamani:
From https://eaflood.atlassian.net/browse/RUBY-413

This adds a rake task and a service object to deal with setting the status of a registration exemption to expired when it is in an active status and the expiry date is in the past.
After talking about expiring on the day of expiring or the day before, we have decided that the registration is considered expired after the date has passed, hence this include a fix on our logic for "not_expired" registrations too.
One more thing part of this PR is the definition of the RegistrationExemption table name which allows us to use the AASM default defined scopes.


app/services/expired_registrations_service.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ExpiredRegistrationsService < ::WasteExemptionsEngine::BaseService
+  def run
+    all_expired_registration_exemptions.update_all(state: :expired)
```
irisfaraway: Can we use `.expire!` to use the state machine instead of directly updating the value in the DB?

See https://github.com/DEFRA/waste-exemptions-back-office-ta/blob/master/app/models/concerns/can_deactivate_exemption.rb#L27-L29

app/models/waste_exemptions_engine/registration_exemption.rb
```diff
@@ -16,7 +18,8 @@ class RegistrationExemption < ActiveRecord::Base
         .order(:registered_on, :registration_id)
     }
 
-    scope :not_expired, -> { where("expires_on > CURRENT_DATE") }
+    scope :not_expired, -> { where("expires_on >= CURRENT_DATE") }
+    scope :expired, -> { where("expires_on < CURRENT_DATE") }
```
irisfaraway: I'm surprised this doesn't clash with any scoping already provided by AASM, since `expired` is also the name of a state. I think something like `should_be_expired` or `past_expiry_date` might be clearer that it's not based on the actual state of the exemption.

irisfaraway: Also, do we need any tests for these? Maybe overkill but let me know what you think.

cintamani: Well, the scope from AASM is defined above, hence this will take precedence. I can add test coverage but I am quite sure. Maybe better to name this differently?

Cruikshanks:
I would like to see https://eaflood.atlassian.net/browse/RUBY-455 get decided on as a priority before we approve and merge this.

Essentially I disagree with the current thinking that an exemption is still ACTIVE on its expiry date, as this

- contradicts what we seem to say to user in our content
- would mean Waste exemptions definition of expires_on differs from Waste Carriers, and I think they should be the same

So I'd like to see this bottomed out before add more code that depends on this decision.

cintamani:
It totally makes sense. Will park this for now until the next session with the team.

app/models/waste_exemptions_engine/registration_exemption.rb
cintamani: On second thoughts, I think is totally fine that we rename it since the status change happens in a background job. It does mean that we will never check the "state" but rather the date, and that's totally fine. Probably then we should move this scope in the same concern that deals with adding AASM. Let me know what you think :)

cintamani:
We discussed and agreed that the rules for "expired" and "not_expired" in this PR are good to go. We will work around the issue of EPR export by scheduling the job at the correct time for the live release. Open issue is around the overriding of the `expired` scope.

app/models/waste_exemptions_engine/registration_exemption.rb
irisfaraway: @cintamani Sounds like a good spot for it to me 👍

cintamani: I was actually suggesting not to rename it and move it in the AASM file for more clarity around the fact that we are overriding it :) Shall we have a call and discuss it later? :)

cintamani:
Merging this as the version that will go out with the release have already been tagged :)

PR #256, opened 2019-07-02
Schedule the registration exemptions expiry job to run daily

cintamani:
From https://eaflood.atlassian.net/browse/RUBY-413

Add configuration and scheduling of the registration exemptions expiry job


.env.example
```diff
@@ -26,6 +26,9 @@ EMAIL_SERVICE_EMAIL=""
 FIRST_RENEWAL_EMAIL_REMINDER_DAYS=28
 FIRST_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME="1:05"
 
+# Expire registration scheduler configs
+EXPIRE_REGISTRATION_EXEMPTION_RUN_TIME="23:00"
```
Cruikshanks: We might as well go with what we'll be setting it to

```suggestion
EXPIRE_REGISTRATION_EXEMPTION_RUN_TIME="00:05"
```

.travis.yml
```diff
@@ -15,6 +15,7 @@ env:
     - EXPORT_SERVICE_BULK_EXPORT_FREQUENCY="friday"
     - EXPORT_SERVICE_BULK_EXPORT_TIME="16:05"
     - FIRST_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME="16:15"
+    - EXPIRE_REGISTRATION_EXEMPTION_RUN_TIME="23:00"
```
Cruikshanks: Unless there is a reason not to, we might as well go with what we'll actually be setting it to

```suggestion
    - EXPIRE_REGISTRATION_EXEMPTION_RUN_TIME="00:05"
```

config/schedule.rb
```diff
@@ -35,3 +35,11 @@
 every :day, at: (ENV["FIRST_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME"] || "1:05"), roles: [:db] do
   rake "email:renew_reminder:first:send"
 end
+
+# This is the registration exemptions exiry job which will collect all active
+# registration exemptions that have an expire date in the past and will set their
+# state to `expired`
+bulk_time = (ENV["EXPIRE_REGISTRATION_EXEMPTION_RUN_TIME"] || "20:05")
```
Cruikshanks: Again, just for consistency we might as well keep this at the time we are going to run the job

```suggestion
bulk_time = (ENV["EXPIRE_REGISTRATION_EXEMPTION_RUN_TIME"] || "00:05")
```

Cruikshanks:
@irisfaraway One for our dev meetings I think about how we define what needs a feature flag. But I'd like to keep them to a minimum. So I'm currently going with

- we've implemented the code but the 'switch on' needs to happen much later and we have other things we want to release
- the functionality is part of a wider feature, and we don't want to enable it till everything related to the feature is done

In the case of this we want it running asap, and everything is good to go, so I don't think it needs a feature flag.

PR #260, opened 2019-07-04
Fix minor bugs in first email reminder

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-252

This change removes an extra border appearing in some versions of Outlook, and removes an unwanted space in a mailto link.


irisfaraway:
Cleared merging this with Andrew as it won't affect the upcoming release.

PR #266, opened 2019-07-09
 Attach logo to email instead of linking to gov.uk 

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-252

Attach logo to email as per WCR instead of linking to Gov.uk

### DIFF
--- a/spec/mailers/renewal_mailer_spec.rb
+++ b/spec/mailers/renewal_mailer_spec.rb
@@ -26,27 +26,27 @@ RSpec.describe RenewalReminderMailer, type: :mailer do
     end
 
     it "includes the correct template in the body" do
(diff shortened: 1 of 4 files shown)


spec/mailers/renewal_mailer_spec.rb
```diff
@@ -26,27 +26,27 @@
     end
 
     it "includes the correct template in the body" do
-      expect(mail.body.encoded).to include("Until our online renewal service is launched, you’ll need to register your exemptions again.")
+      expect(mail.body.parts[0].body.encoded).to include("Until our online renewal service is launched, you’ll need to register your exemptions again.")
```
cintamani: Adding an attachment cause the email to become a multipart email, and in order to get the correct part body, this addition was necessary.

PR #276, opened 2019-07-12
Generate Boxi report CSV files for AWS export

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-460

Add a rake task to generate reports for boxy. 
5 tables have to be exported in CSV file, completed of all records and with the same columns present in the SQL structure. 
The code will dynamically generate the correct CSV files based on the records and table structure at the moment of run. 
A main generator service function as an entry point and is responsible mainly for creating the zip file content and load it to the correct AWS bucket. 
Serializers have been created to deal with the generation of the correct CSV content based on the structure of the target table. 
Integration test coverage has been added at the main entry point level. 

Related: 
- Schedule the job via Cron scheduler: https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/277
- Sort out jenkinsfile for QA purposes: https://gitlab-dev.aws-int.defra.cloud/open/rails-deployment/merge_requests/18


app/services/boxi_export/generator_service.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+require "zip"
+require_relative "../concerns/can_load_file_to_aws"
+
+module
+BoxiExport
```
irisfaraway: Accidental line break?

config/application.rb
```diff
@@ -59,6 +59,10 @@ class Application < Rails::Application
 
     config.years_before_expiry = ENV["YEARS_BEFORE_EXPIRY"] || 3
 
+    # BOXY data export
```
irisfaraway:
```suggestion
    # BOXI data export
```

.env.example
```diff
@@ -44,6 +44,11 @@ AWS_DAILY_EXPORT_ACCESS_KEY_ID=<key_id>
 AWS_DAILY_EXPORT_SECRET_ACCESS_KEY=<secret_key>
 AWS_DAILY_EXPORT_BUCKET=<bucket_name>
 
+AWS_BOXI_EXPORTS_BUCKET=<bucket_name>
+AWS_DAILY_EXPORT_ACCESS_KEY_ID=<key_id>
```
irisfaraway:
```suggestion
AWS_BOXI_EXPORT_ACCESS_KEY_ID=<key_id>
```

.env.example
```diff
@@ -44,6 +44,11 @@ AWS_DAILY_EXPORT_ACCESS_KEY_ID=<key_id>
...
+AWS_DAILY_EXPORT_SECRET_ACCESS_KEY=<secret_key>
```
irisfaraway:
```suggestion
AWS_BOXI_EXPORT_SECRET_ACCESS_KEY=<secret_key>
```

.env.example
irisfaraway: Although having made this suggestion – what are we using this env var for? Can't see it in use but maybe I've missed something obvious!

app/services/boxi_export/generator_service.rb
cintamani: yep 😮

.env.example
cintamani: No, you are not. I have missed adding that file in. Is the config for AWS gem!

PR #283, opened 2019-07-26
Add feature toggle on Boxi export

cintamani:
Add a feature toggle around the boxi export, as that's not ready to be released yet.

### DIFF
--- a/lib/tasks/boxi_export.rake
+++ b/lib/tasks/boxi_export.rake
@@ -3,6 +3,6 @@
 namespace :boxi_export do
   desc "Generates a Zip file containing data for BOXI export and load them to AWS bucket"
   task generate: :environment do
-    BoxiExport::GeneratorService.run
+    BoxiExport::GeneratorService.run unless WasteExemptionsEngine::FeatureToggle.active?(:generate_boxi_report)
   end
 end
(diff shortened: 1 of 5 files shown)


irisfaraway (commented):
Can I suggest putting the check on the rake task instead? This is what we did for the emails so it'd be more consistent, and saves the service from being called at all. Also means we don't need to modify the test.
https://github.com/DEFRA/waste-exemptions-back-office-ta/blob/master/lib/tasks/email.rake#L13

lib/tasks/boxi_export.rake
```diff
@@ -3,6 +3,6 @@
 namespace :boxi_export do
   desc "Generates a Zip file containing data for BOXI export and load them to AWS bucket"
   task generate: :environment do
-    BoxiExport::GeneratorService.run
+    BoxiExport::GeneratorService.run unless WasteExemptionsEngine::FeatureToggle.active?(:generate_boxi_report)
```
irisfaraway:
```suggestion
    BoxiExport::GeneratorService.run if WasteExemptionsEngine::FeatureToggle.active?(:generate_boxi_report)
```

irisfaraway (changes requested):

irisfaraway (approved):

PR #287, opened 2019-07-31
Move BOXI export to reports namespace

Cruikshanks:
Spotted that the BOXI export was in its own namespace, and using its on env vars for config. On reflection as

- its an automated job expected to be run nightly
- it exports a bunch of data as CSV files and then them to AWS S3

It was felt it could be moved into the reports namespace to make things consistent. It does compress the csv files first before loading, but that is the only real difference.

So this change updates the code to move the boxi export logic to the `reports` namespace, and updates the naming of it and its env vars to be consistent with the other export functions.


app/services/reports/boxi_export_service.rb
```diff
@@ -3,17 +3,17 @@
...
     include CanLoadFileToAws
 
     def run
       Dir.mktmpdir do |dir_path|
-        BoxiExport::AddressesSerializer.export_to_file(dir_path)
-        BoxiExport::ExemptionsSerializer.export_to_file(dir_path)
-        BoxiExport::PeopleSerializer.export_to_file(dir_path)
-        BoxiExport::RegistrationExemptionsSerializer.export_to_file(dir_path)
-        BoxiExport::RegistrationsSerializer.export_to_file(dir_path)
+        Boxi::AddressesSerializer.export_to_file(dir_path)
```
cintamani, line 12: Like this! :D

cintamani (approved):
Nice! It does look better organised like this :) Thanks, Alan :D

PR #290, opened 2019-07-31
Update renewal email to use magic link

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-492

Now that we have created the magic link route in https://github.com/DEFRA/waste-exemptions-engine/pull/245, we need to actually add it to the emails.

This will also require a feature flag so we don't deploy this version of the email until the whole renewals process is ready to release.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -64,9 +64,6 @@ module WasteExemptionsBackOffice
     config.email_test_address = ENV["EMAIL_TEST_ADDRESS"]
     config.first_renewal_email_reminder_days = ENV["FIRST_RENEWAL_EMAIL_REMINDER_DAYS"]
 
-    # Renewal
-    config.registration_renewal_grace_window = ENV["REGISTRATION_RENEWAL_GRACE_WINDOW"] || 30
-
     # Version info
     config.application_name = "waste-exemptions-back-office-ta"
     config.git_repository_url = "https://github.com/DEFRA/#{config.application_name}"
--- a/config/application.rb
+++ b/config/application.rb
@@ -64,6 +64,9 @@ module WasteExemptionsBackOffice
     config.email_test_address = ENV["EMAIL_TEST_ADDRESS"]
     config.first_renewal_email_reminder_days = ENV["FIRST_RENEWAL_EMAIL_REMINDER_DAYS"]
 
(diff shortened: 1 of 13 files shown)


config/application.rb
```diff
@@ -64,6 +64,9 @@ class Application < Rails::Application
     config.email_test_address = ENV["EMAIL_TEST_ADDRESS"]
     config.first_renewal_email_reminder_days = ENV["FIRST_RENEWAL_EMAIL_REMINDER_DAYS"]
 
+    # Renewal
+    config.registration_renewal_grace_window = ENV["REGISTRATION_RENEWAL_GRACE_WINDOW"] || 30
```
cintamani: We merged this fix already, so this is a duplicate now :) https://github.com/DEFRA/waste-exemptions-back-office-ta/blob/3f41bb9b337d45873971d953cdf6a8bdbd5efa68/config/application.rb#L76

db/schema.rb
```diff
@@ -195,7 +195,7 @@
     t.boolean  "address_finder_error",   default: false
     t.datetime "created_at",                             null: false
     t.datetime "updated_at",                             null: false
-    t.string   "type",                                   null: false
+    t.string   "type"
```
cintamani: I think this is coming from your local and shouldn't be merged

lib/tasks/one_off/reminders.rake
```diff
@@ -6,4 +6,10 @@ namespace :reminder do
     registration = WasteExemptionsEngine::Registration.last
     RenewalReminderMailer.first_reminder_email(registration).deliver_now
   end
+
+  desc "Send an example first magic link email"
+  task magic_link: :environment do
+    registration = WasteExemptionsEngine::Registration.last
+    RenewalReminderMailer.first_renew_with_magic_link_email(registration).deliver_now
```
cintamani, line 13: ❤️

cintamani (changes requested):
A couple of little things :) Looks good

config/application.rb
irisfaraway: Oops. A consequence of the rebase! Fixed in deb423c

cintamani (approved):
🐑 🇮🇹

PR #293, opened 2019-07-31
Add configs for generating and decoding a renew JWT token

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-514

Add configs to the application to being able to generate and decode correctly a JWT token.


config/application.rb
```diff
@@ -67,5 +67,8 @@ class Application < Rails::Application
     # Version info
     config.application_name = "waste-exemptions-back-office-ta"
     config.git_repository_url = "https://github.com/DEFRA/#{config.application_name}"
+
+    # Change authomatic expire of renew's magic link token
```
Cruikshanks: Not sure this comment is applicable here. I think it would be better to use the same comment as in `.env.example` as the basis for one here.

```suggestion
    # Grace window value in days from when a registration expires. During the
    # grace window users will still be permitted to renew their registration.
```

Cruikshanks (changes requested):
Just one request to update one of the comments. Other than that all good!

config/application.rb
cintamani: Yes! this was an unfortunate copy-paste :D

Cruikshanks (approved):

PR #306, opened 2019-08-02
Use ENV variable to setup magic link feature toggle

cintamani:


irisfaraway:
@cintamani Should we also change the tests to stub the env var instead of the response from the feature toggle?

cintamani:
@irisfaraway I don't think that's necessary at all. All Feature toggles behaviour is well tested in the engine

irisfaraway (approved):

PR #316, opened 2019-08-07
Add renew link in back office dashboard

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-499

Add links to renew journey on all registration in the renewal period.
The link will redirect first to the ad_privacy_policy and then to the renewal journey.
Only users with at least admin permissions can see and use the path.

<details> 
<summary>Before renewal period:</summary>

Screenshot 2019-08-07 at 12 39 13

</details>
<details> 
<summary>In renewal period:</summary>

Screenshot 2019-08-07 at 12 39 02

</details>

<details> 
<summary>Past renewal period:</summary>

Screenshot 2019-08-07 at 12 38 56

</details>

### DIFF
--- a/app/views/shared/_search_result.html.erb
+++ b/app/views/shared/_search_result.html.erb
@@ -44,7 +44,7 @@
         <% end %>
       </ul>
     <% else %>
-      <br />
+      <br>
     <% end %>
   </div>
 
--- a/app/views/shared/_search_result.html.erb
+++ b/app/views/shared/_search_result.html.erb
@@ -6,17 +6,19 @@
         <span><%= t(".labels.status") %></span>
         <span class="<%= "status-tag-#{status_tag_for(result)}" %>"><%= t(".statuses.#{status_tag_for(result)}") %></span>
       </li>
+
       <% if result.operator_name.present? %>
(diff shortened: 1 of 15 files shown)


app/views/shared/_search_result.html.erb
```diff
@@ -6,67 +6,73 @@
...
-        </li>
+          <li>
+            <span><%= t(".labels.grid_reference") %></span>
+            <%= result.site_address.grid_reference %>
+          </li>
         <% end %>
       </ul>
     <% else %>
-      <br>
+      <br />
```
Cruikshanks: Minor tweak, and though valid its our understanding that void elements should just be defined using the opening tag syntax

```suggestion
      <br>
```

See
- <https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br>
- <https://dev.w3.org/html5/html-author/#void>
- <https://stackoverflow.com/a/1946446>

Cruikshanks (changes requested):
Its the nit-pickiest of changes, but it is only that I can see 😁

Cruikshanks (approved):

PR #324, opened 2019-08-09
Add anonymise emails rake task

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-461

Add a rake task to clean up email addresses from Registration and TransientRegistration objects
See also: https://gitlab-dev.aws-int.defra.cloud/open/rails-deployment/merge_requests/21


irisfaraway (dismissed):
The A/C says that rather than replacing all email addresses with the same values, we should change the domain: "all emails have the domain element replaced with `@example.com".

I assume this is so we don't just email the same address loads of times.

It'll also let us simulate having a different applicant and contact email.

lib/tasks/anonymise_emails.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+desc "Set all email domain addresses to given address or default one." \
+      " Usage: `rake anonymise_emails TEST_DOMAIN=test@testmedefra.gov.uk`"
+task anonymise_emails: :environment do
+  test_email = ENV["TEST_EMAIL"].presence || "test@testmedefra.gov.uk"
```
cintamani:
```suggestion
  test_email = ENV["TEST_EMAIL"].presence || "test@example.com"
```

lib/tasks/anonymise_emails.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+desc "Set all email domain addresses to given address or default one." \
+      " Usage: `rake anonymise_emails TEST_DOMAIN=test@testmedefra.gov.uk`"
+task anonymise_emails: :environment do
```
Cruikshanks: I would prefer to move this to the `lib/tasks/email.rake` task so that it comes under the `email` namespace eg `bundle exec rake email:anonymise`.

Do you have any issues with that @cintamani ?

cintamani: Nope!! Sounds good! :)

Cruikshanks (dismissed):

Cruikshanks (approved):

PR #327, opened 2019-08-09
Add ability to custom-parse attribute on Boxi serializer

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-556

We had the need to strip out newlines on multiline params such as `description` in the `addresses` table.
To do so, I have used some methaprogramming to create parse methods for an attribute so that we can easely twick those value in the future.


Cruikshanks:
Just shows what an ace developer @cintamani is! I forgot to say we have the issue in another field, but thanks to how the problem has been solved that won't be a problem ❤️ 🥇 😁

Cruikshanks (approved):

PR #332, opened 2019-08-12
Add second email reminder for renewals

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-253

Add second renewal email reminder with magic link
Screenshot 2019-08-12 at 12 24 57


cintamani:
I am not going to add test coverage for those private methods Codeclimate is complaining about. 
Reasons: 
 - Those are private methods.
 - Any class implementing the base class will have its own test coverage which will fail if those methods aren't implemented.

app/services/second_renewal_reminder_service.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class SecondRenewalReminderService < RenewalReminderServiceBase
+
+  private
+
+  def send_email(registration)
+    RenewalReminderMailer.second_reminder_email(registration).deliver_now
```
Cruikshanks, line 7: The first email reminder is controlled from a feature toggle. The second should be as well (at least initially) particularly as we have told NCCC that it can be switched off with a simple environment config change.

I think in the next release we can start to look at removing the feature toggles of things that no longer require them.

config/schedule.rb
```diff
@@ -36,6 +36,13 @@
   rake "email:renew_reminder:first:send"
 end
 
+# This is the daily second renewal reminder mail service.
+# Will run once a day in the early morning hours and send email reminders about
+# registrations that will expire in X time.
+every :day, at: (ENV["SECOND_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME"] || "1:05"), roles: [:db] do
```
Cruikshanks: Can we have the default time set to 03:05? This is a time when no other jobs will be running.

lib/tasks/email.rake
```diff
@@ -26,5 +27,15 @@ namespace :email do
         Airbrake.close
       end
     end
+
+    namespace :second do
+      desc "Collect all registration that expires in X weeks and sends a second email reminder"
```
Cruikshanks: I think it will be clearer to say

```suggestion
      desc "Send second email reminder to all registrations expiring in X days (default is 14)"
```

I know this does not match the description for the first one. But it also talks in weeks when we set the value in days. So I'd also suggest / be happy with the description for the first reminder task to be updated as well in this PR to

```ruby
      desc "Send first email reminder to all registrations expiring in X days (default is 28)"
```

config/application.rb
```diff
@@ -62,6 +62,7 @@ class Application < Rails::Application
 
     # Emails
     config.email_test_address = ENV["EMAIL_TEST_ADDRESS"]
+    config.second_renewal_email_reminder_days = ENV["SECOND_RENEWAL_EMAIL_BEFORE_DAYS"]
```
Cruikshanks: We have a default value set for the first email reminder and the grace window. I think we also need a default value set here of `14`.

```suggestion
    config.second_renewal_email_reminder_days = ENV["SECOND_RENEWAL_EMAIL_BEFORE_DAYS"] || 14
```

When we release we will be overriding this to `7`, but we intend to return to the default of 14 days once the registration peak is at an end.

app/services/renewal_reminder_service_base.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  private
+
+  def send_email
+    raise(NotImplementedError)
+  end
+
+  def expiring_registrations
```
Cruikshanks, line 21: This handles the criteria for the first reminder email. The problem is the A/C for the second reminder email has this

> If the registration triggering the email appears to be already renewed via Magic link, this email won’t be triggered to be sent.

So we somehow need to cover this difference.

Cruikshanks (dismissed):
I've requested a number of changes. Shout if any don't make sense or you disagree with them.

config/application.rb
cintamani: I don't see any default for the first email reminder and grace window 🤔 will update this anyway

app/services/second_renewal_reminder_service.rb
cintamani: Added in the rake task as per first email reminder.

config/application.rb
cintamani: Oh. I rebased XD

config/schedule.rb
```diff
@@ -36,6 +36,13 @@
   rake "email:renew_reminder:first:send"
 end
 
+# This is the daily second renewal reminder mail service.
+# Will run once a day in the early morning hours and send email reminders about
+# registrations that will expire in X time.
+every :day, at: (ENV["SECOND_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME"] || "3:05"), roles: [:db] do
```
Cruikshanks: 🤦‍♂ 🤦‍♂ 🤦‍♂ 🤦‍♂ 🤦‍♂ 🤦‍♂ 

```suggestion
every :day, at: (ENV["SECOND_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME"] || "04:05"), roles: [:db] do
```

Spotted we've defaulted the BOXI export at 3am 😳

spec/models/waste_exemptions_engine/registration_spec.rb
```diff
@@ -33,6 +33,18 @@
     end
   end
 
+  describe ".renewals" do
+    it "retruns registration that are renewals of older registrations" do
```
Cruikshanks: Just a typo

```suggestion
    it "returns registrations that are renewals of older registrations" do
```

.env.example
```diff
@@ -23,7 +23,9 @@ EMAIL_TEST_ADDRESS=""
 EMAIL_SERVICE_EMAIL=""
 
 # Renewal emails config
+SECOND_RENEWAL_EMAIL_BEFORE_DAYS=14
 FIRST_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME="1:05"
+SECOND_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME="1:05"
```
Cruikshanks: See later comment where I have screwed up the time (🤦‍♂) but our default for this job is now "04:05" in the schedule, so would be great to update the `.env.example` to match.

Cruikshanks (changes requested):
Very minor changes, 2 of which are my fault! 😁 Other than that this is great!

config/schedule.rb
cintamani: I should have spotted it myself :D

.env.example
cintamani:
```suggestion
SECOND_RENEWAL_EMAIL_REMINDER_DAILY_RUN_TIME="04:05"
```

Cruikshanks (approved):

PR #334, opened 2019-08-13
Refactor email reminders to remove non-magic link

Cruikshanks:
The original plan was to ship a version of the first reminder email that only asked users to re-register. This was because it was thought shipping renewals would take some time.

Thanks to the success of the team, we actually have renewals including the version of the email which includes a **magic link** ready to go. This means the old re-register email will never see the light of day.

However we still have code that makes distinctions between the re-register and magic link versions of the first email reminder. So this change is some housekeeping to clean up the logic around the first email reminder to remove those distinctions.

<details>
<summary>First reminder</summary>
Screenshot 2019-08-13 at 17 10 07

</details>

<details>
<summary>Second reminder</summary>
Screenshot 2019-08-13 at 17 11 13

</details>


Cruikshanks:
Sorry, jumped the gun a bit. Should check the emails are actually ok before asking for a review!

app/mailers/renewal_reminder_mailer.rb
```diff
@@ -7,11 +7,12 @@ class RenewalReminderMailer < ActionMailer::Base
 
   def first_reminder_email(registration)
     assign_values_for_email(registration)
+    @magic_link_url = magic_link_url(registration)
```
cintamani: Can we move this inside the `assign_values_for_email` method?

spec/mailers/renewal_mailer_spec.rb
```diff
@@ -72,11 +73,16 @@
...
     end
+
+    it "includes the correct renewal link" do
+      url = "https://wex.gov.uk/renew/#{registration.renew_token}"
+      expect(mail.body.parts[0].body.encoded).to include(url)
+    end
   end
 
-  describe "first_renew_with_magic_link_email" do
+  describe "second_reminder_email" do
```
cintamani, line 83: I missed this! Thanks!

cintamani (changes requested):

Cruikshanks:
Ok, checked the emails now. As far as I can see I haven't mucked anything up, so this is good to go for a review (thanks for the initial heads up @cintamani 😁 )

cintamani (approved):
:)

PR #335, opened 2019-08-13
Add links to resend renew email in back office

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-466

Add links in the BO dashboard and resource details page to resend a renewal email.
Add style of error and message flashes + partials.

<details> 
<summary> Screenshots: </summary>

Screenshot 2019-08-13 at 16 48 31
Screenshot 2019-08-13 at 16 51 30
Screenshot 2019-08-13 at 16 59 56
Screenshot 2019-08-13 at 17 00 32

</details>


spec/requests/resend_renewal_email_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        expect(response.code).to eq("302")
+      end
+
+      it "return a success message" do
+        success_message = I18n.t("resend_renewal_email.messages.success", email: registration.contact_email)
+
+        get request_path, {}, "HTTP_REFERER" => "/"
+        follow_redirect!
+
+        # binding.pry
```
cintamani:
```suggestion
```

spec/requests/resend_renewal_email_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+        around do |example|
+          # We want bullet disable for eager loading.
+          # We are not concerned about an unused eager loading when an error happens
+          Bullet.unused_eager_loading_enable = false
+
+          example.run
+
+          Bullet.unused_eager_loading_enable = true
+        end
```
cintamani, line 55: @irisfaraway ☝️ nice little trick to disable Bullet when he is unreasonable :D

app/controllers/resend_renewal_email_controller.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class ResendRenewalEmailController < ApplicationController
+  def new
+    authorize
+
+    begin
+      RenewalReminderMailer.first_renew_with_magic_link_email(registration).deliver_now
```
cintamani:
```suggestion
      RenewalReminderMailer.first_reminder_email(registration).deliver_now
```

spec/requests/resend_renewal_email_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+        get request_path, {}, "HTTP_REFERER" => "/"
+        follow_redirect!
+
+        expect(response.body).to include(success_message)
+      end
+
+      context "when an error happens", disable_bullet: true do
+        before do
+          expect(RenewalReminderMailer).to receive(:first_renew_with_magic_link_email).and_raise(StandardError)
```
cintamani:
```suggestion
          expect(RenewalReminderMailer).to receive(:first_reminder_email).and_raise(StandardError)
```

app/assets/stylesheets/partials/_flash_messages.scss
```diff
@@ -0,0 +1,16 @@
+.error-summary.wex_message {
```
Cruikshanks: It's going to have a knock on effect throughout the PR, but can we rename this as `flash_message` or something else more generic?

It is highly likely we will want to use this same decision and functionality in WCR renewals, so dropping the service specific prefix will make it easier to port.

app/controllers/resend_renewal_email_controller.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class ResendRenewalEmailController < ApplicationController
+  def new
+    authorize
+
+    begin
+      RenewalReminderMailer.first_reminder_email(registration).deliver_now
+
+      flash[:message] = I18n.t("resend_renewal_email.messages.success", email: registration.contact_email)
```
Cruikshanks: With this merged its going to make implementing [RUBY-467 Add option to resend confirmation email](https://eaflood.atlassian.net/browse/RUBY-467) much easier as well 🎉 😁 

So on the basis we know this other story is needed, and this is a pattern we'll want to use in WCR, is there a way we can move the implementation for assigning `flash[:message]`, and `flash[:error]` somewhere else? A controller concern for example?

Then rather than duplicating this across controllers, we can just call something like

```ruby
flash_success("Email sent. Well done you!")
flash_error("You broke it", "You appear to have tried using the system and broke it. Someone's in trouble!")

# ...

def flash_success(message)
  flash[:message]
end

def flash_error(message, description)
  flash[:error] = error
  flash[:error_details] = description
end
```

It is pre-emptive refactoring, but I'm 99.999% sure this will be reused!

Cruikshanks (changes requested):
Left some specific comments. But essentially this PR not only adds the resend renewal link, it implements in-page confirmation messages for the first time in one of our back office apps.

I'm 100% certain it will be reused, not just in WEX but in WCR as well. As such I'm looking at this implementation with an eye to immediate reuse, hence my comments.

Ping me if you want to discuss this further. But other than that, this is awesome!

app/controllers/resend_renewal_email_controller.rb
cintamani: Sure! Will do that :)

app/assets/stylesheets/partials/_flash_messages.scss
cintamani: Martin provided me with style and HTML for the messages. 
The main issue is that we are reusing style from the existing `error-summary` classes, which are defined in the Elements css, and give us responsiveness for free. This extra code here is just to adapt them to look as per screenshots. 
My suggestion is that we leave this as-it-is here and if this becomes an "important" bit later on before we move to the new GDS, then its worth spending more time tweaking this style.

Cruikshanks: I'm fine with the styles, how it works etc. It's a great idea to reuse what we already have. I just don't want to use the prefix `wex` in this code.

Cruikshanks:
@cintamani As we're all keen to get this into the next release I went ahead and committed the changes I was looking to see, plus a couple of fixes we spotted (extra space and the `flash_error()` error concern).

As you can see regards the CSS styles thing, I wasn't looking to refactor the code, simply swap the `wex` prefix for something more generic. Completely agree we should hold off investing too much time into it at this point until we have a firmer idea about implementing the design system.

Anyway, hope I've not stomped on your toes 😰 😬 !

irisfaraway (approved):
❤️ And flash messages too! Woohoo!

Cruikshanks (approved):

cintamani:
Not at all!! Glad you did it actually :) Thanks :)

PR #349, opened 2019-08-22
Make the registration responsible to know if it is renewable

cintamani:
This is a little refactoring on the action links helpers for renewal links.
First, we drop the check on the class name as the `can?` permission check will just fail unless the object is of the right class.
Second, we give responsibility to check if a registration is renewable to the registration itself.


app/models/waste_exemptions_engine/registration.rb
```diff
@@ -29,5 +33,13 @@ def state
 
       "ceased"
     end
+
+    def renewable?
```
Cruikshanks, line 37: Though minor these new methods are public. So do you think we should have some unit tests for them?

cintamani: Sure! will add some

spec/models/waste_exemptions_engine/registration_spec.rb
```diff
@@ -45,45 +19,62 @@
...
+  end
+
+  describe "#in_renewable_state?" do
+    before do
+      allow(registration).to receive(:state).and_return(state)
+    end
 
-    context "when the renewal period has started" do
-      let(:expires_on) { 10.days.from_now }
+    context "whent the state is active" do
```
irisfaraway:
```suggestion
    context "when the state is active" do
```

spec/models/waste_exemptions_engine/registration_spec.rb
```diff
@@ -45,45 +19,62 @@
...
+    context "when the registration is not in a renewal window" do
+      it "returns false" do
+        allow(registration).to receive(:in_renewal_window?).and_return(false)
+        allow(registration).to receive(:in_renewable_state?).and_return(true)
+
+        expect(registration).to_not be_renewable
+      end
+    end
 
+    context "when the registratiion is not in a renewal state" do
```
irisfaraway:
```suggestion
    context "when the registration is not in a renewal state" do
```

irisfaraway (changes requested):
Looks good :) Just some typos!

irisfaraway (approved):

PR #350, opened 2019-08-22
Add rake task to remove old transient_registrations

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-326

We've agreed that we shouldn't store transient_registrations for more than 30 days. So we need a way of clearing out the old ones.

This rake task will find and remove transient_registrations that are older than 30 days. It should also remove all related objects (eg transient_addresses).

### DIFF
--- a/app/services/transient_registration_cleanup_service.rb
+++ b/app/services/transient_registration_cleanup_service.rb
@@ -2,7 +2,7 @@
 
 class TransientRegistrationCleanupService < ::WasteExemptionsEngine::BaseService
   def run
-    transient_registrations_to_remove.each(&:destroy)
+    transient_registrations_to_remove.destroy_all
   end
 
   private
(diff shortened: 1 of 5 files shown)


irisfaraway:
Planning to add the scheduling separately.

app/services/transient_registration_cleanup_service.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class TransientRegistrationCleanupService < ::WasteExemptionsEngine::BaseService
+  def run
+    transient_registrations_to_remove.each(&:destroy)
```
Cruikshanks: I've not tested directly what the performance implications are, but I have tested locally and this works in the same way

```suggestion
    transient_registrations_to_remove.destroy_all
```

I'm assuming that this would be more performant that iterating through each result and calling `destroy()`. Plus I think the intent is slightly more clearer. What do you think?

Cruikshanks (approved):

PR #355, opened 2019-08-27
Show an already renewed message when relevant

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-602

Show an already renewed message in the back office instead of the renew links as per business rules.

Screenshot 2019-08-27 at 10 39 32
Screenshot 2019-08-27 at 10 39 25

### DIFF
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -57,6 +57,12 @@ module ActionLinksHelper
       resource_has_active_or_expired_exemptions?(resource)
   end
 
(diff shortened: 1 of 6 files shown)


Cruikshanks (changes requested):
We've now added unit tests in for the other action view helpers. So I think we should also do that for this one if that's ok.

cintamani:
I swear I wrote it 😮 🤦‍♀ 

_franctically typing a spec_

Cruikshanks (approved):

PR #356, opened 2019-08-27
Add link to renewed from registration in details page

cintamani:
Closes: https://eaflood.atlassian.net/browse/RUBY-506

Add link to the referring registration in the resource details page

Screenshot 2019-08-27 at 11 05 56


irisfaraway:
Looks like the build is broken. I think because the `resource` in this case can be a Registration or a TransientRegistration, which doesn't have a `referring_registration` relationship.

cintamani:
Oh!! Now it makes sense!! Thanks!

irisfaraway (dismissed):
Let me know if you need a re-approval after merging the engine update 👍

irisfaraway (approved):

PR #357, opened 2019-08-27
Generate assisted digital renewal letter

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-578

For users who didn't register with an email address, we want to notify them by post.

This means we'll need to generate a PDF of the letter as part of the back office.

This PR implements the PDF generation and adds a way to test it (pending adding the proper ways to access it).


app/presenters/confirmation_letter_presenter.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class ConfirmationLetterPresenter < BasePresenter
+class ConfirmationLetterPresenter < BaseLetterPresenter
```
cintamani, line 3: <3

app/presenters/base_letter_presenter.rb
```diff
@@ -0,0 +1,36 @@
...
+    address_lines(operator_address).join(", ")
+  end
+
+  def site_address_one_liner
+    address_lines(site_address).join(", ")
+  end
+
+  private
+
+  def format_name(first_name, last_name)
```
cintamani: What about including the `waste_exemptions_engine/application_helper` here and use the implementation of this method from it?

config/routes.rb
```diff
@@ -25,6 +25,10 @@
 
   get "/confirmation-letter/:id", to: "confirmation_letter#show", as: :confirmation_letter
 
+  # Renewal Letter
+
```
cintamani:
```suggestion
```

spec/support/shared_examples/presenters/a_letter_presenter.rb
```diff
@@ -0,0 +1,86 @@
...
+      it "returns a string representation of the address" do
+        address = registration.site_address
+        address_fields = [
+          address.organisation,
+          address.premises,
+          address.street_address,
+          address.locality,
+          address.city,
+          address.postcode
+        ].reject(&:blank?)
```
cintamani: Do we need this `blank?` check? Feels strange to have it in a test, we should control the data we use rather than manipulate them if that makes sense :)

cintamani (changes requested):
A couple of minor suggestions :)

config/routes.rb
irisfaraway: Resolved by https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/357/commits/70006d2f56bf3dfdf1325e9a28ba478f2b310357 (had to make more changes for this to be consistent)

app/presenters/base_letter_presenter.rb
irisfaraway: Resolved by https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/357/commits/75412f2a638796b5192063b3ebd8505c92f459b2

spec/support/shared_examples/presenters/a_letter_presenter.rb
irisfaraway: Resolved by https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/357/commits/b120cf0d51c9a5899429642d2fd83fe0caa9b636

irisfaraway:
@cintamani Ready for another look!

app/presenters/renewal_letter_presenter.rb
```diff
@@ -0,0 +1,69 @@
...
+    @_listable_exemptions ||= exemptions.first(18)
+    @_listable_exemptions
+  end
+
+  def unlisted_exemption_count
+    @_unlisted_exemption_count ||= calculate_number_of_unlisted_exemptions
+  end
+
+  def partners_list
+    return unless partnership?
```
cintamani:
```suggestion
```
I don't think we need to do this at all. The `partnership?` check already happen in the view. In the `worst case` it make anyway sense for this method to return an empty array when this is not a partnership rather than `nil`. So that if you expect a method to return an array you can check if the array contains object rather than having to remember that it can also return a `nil` object, which won't respond to methods like `count` and `size`.

app/presenters/renewal_letter_presenter.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    people_list = []
+    people.each do |person|
+      people_list << format_names(person.first_name, person.last_name)
+    end
+
```
cintamani: What about: 
```
people.map do |person|
  format_names(person.first_name, person.last_name)
end.join(", ")
```

app/presenters/renewal_letter_presenter.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+  end
+
+  def site_description_abridged?
+    @_site_description_abridged ||= site_address.description.length > max_site_description_length
+  end
+
+  private
+
+  def max_site_description_length
```
cintamani: I think this is a good candidate for a class constant rather than a private method, WDYT?

app/controllers/renewal_letter_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class RenewalLetterController < ApplicationController
+  def show
+    registration = WasteExemptionsEngine::Registration.find(params[:id])
+    authorize! :read, registration
+
+    show_as_html = params[:show_as_html].present? && params[:show_as_html] == "true"
```
cintamani:
```suggestion
    show_as_html = params[:show_as_html] == "true"
```
I don't think we need the extra `present?` here :)

app/presenters/base_letter_presenter.rb
```diff
@@ -0,0 +1,34 @@
...
+    address_lines(site_address).join(", ")
+  end
+
+  private
+
+  def address_lines(address)
+    return [] unless address
+
+    address_fields = %i[organisation premises street_address locality city postcode]
+    address_fields.map { |field| address.public_send(field) }.reject(&:blank?)
```
cintamani, line 32: [NIT] 
```suggestion
    address_fields.map { |field| address.public_send(field).presence }.compact
```

app/presenters/renewal_letter_presenter.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+class RenewalLetterPresenter < BaseLetterPresenter
+  # Provides the full postal address for the letter.
+  def postal_address_lines
+    lines = [contact_full_name]
+    lines << operator_name unless business_type == WasteExemptionsEngine::Registration::BUSINESS_TYPES[:sole_trader]
```
cintamani, line 9:
```suggestion
    lines << operator_name unless business_type == sole_trader?
```
The `#sole_trader?` method should be implemented in the engine `CanHaveRegistrationAttributes` to be correct so this is a NIT

app/presenters/renewal_letter_presenter.rb
```diff
@@ -0,0 +1,69 @@
...
+    lines << address_lines(contact_address)
+    lines.flatten!.reject(&:blank?)
```
cintamani, line 11: [NIT] this can also benefit from the use of the `presence` and `compact` methods

cintamani (changes requested):
Sorry! I have spotted some more things. Mostly NITs but still good to mention the tricks :D

app/presenters/base_letter_presenter.rb
Cruikshanks: I get what `presence` and `compact` do, but I don't understand why they are better to use than `reject(&:blank?)`.

The key point for me is that the intent is more meaningful using `reject(&:blank?)` whereas I had to go google what presence and compact do. However this is 100% my own opinion. The rest of the team might feel different.

app/presenters/renewal_letter_presenter.rb
Cruikshanks: See my previous comment about replacing `reject()` with `presence` and `compact`.

app/presenters/renewal_letter_presenter.rb
Cruikshanks: ❤️ We might have to work through the impact to ensure we don't break anything but I think @cintamani is right. My bad, think this was my code originally 🤦‍♂ 😳

app/presenters/renewal_letter_presenter.rb
Cruikshanks: ❤️ That's helps skip a few steps!

app/presenters/base_letter_presenter.rb
cintamani: This is why this is an NIT. Me, I have used `presence` and `compact` so many times that I feel that's more clear, this is more of a "this is another way of achieving this", but not necessarily a better one :)

app/presenters/renewal_letter_presenter.rb
irisfaraway: Resolved by https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/357/commits/5f6b9a1ac8b05dd7936852a84bdeb554c9bf2e48

app/presenters/renewal_letter_presenter.rb
irisfaraway: Resolved by https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/357/commits/5f6b9a1ac8b05dd7936852a84bdeb554c9bf2e48

app/presenters/renewal_letter_presenter.rb
irisfaraway: Resolved by https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/357/commits/f156d14ad5cd44ad88c76ac4a9da4000480091b2

app/presenters/renewal_letter_presenter.rb
irisfaraway: I agree, but think I'd rather do this for all business types as a separate PR and then update this once the engine change is rolled out. What do you think?

app/presenters/base_letter_presenter.rb
irisfaraway: `reject(&:blank)` is what I'm most used to and I think we've used it fairly consistently around the various codebases, so I'd rather stick with that. But useful to know about alternatives 👍

app/controllers/renewal_letter_controller.rb
irisfaraway: Resolved by https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/357/commits/5ca1395696143a211817bd537f9f3e9441fbad31 (and then some)

app/presenters/base_letter_presenter.rb
cintamani: Sounds good! :D

app/presenters/renewal_letter_presenter.rb
cintamani: I think it makes sense :)

cintamani (approved):

PR #383, opened 2019-09-10
AD renewal letters bulk generation

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-580 And https://eaflood.atlassian.net/browse/RUBY-581

Add rake task and classes to generate bulk AD renewal letters.
For every AD expiring registration, a letter pdf is created and they are then concatenated in a single PDF with blank pages between the letters.
A refactoring on the CSS that generate the letter was necessary in order for use to get advantage of the wicked PDF HTML to PDF power. Specifically, no `position:absolute` can be used as those gets positioned in the first page always for all letter rather than in their letter's page.
A new table is also added which will store info about the bulk generation. Fields for https://eaflood.atlassian.net/browse/RUBY-582 have been added as well, such as `printed_by` and `printed_on`.
The table will also hold a file name that will be used by the Defra ruby AWS gem to generate the correct download file for the AWS bucket.


spec/services/ad_renewal_letters_export_service_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  describe ".run" do
+    let(:bucket) { double(:bucket) }
+    let(:result) { double(:result, successful?: true) }
+
+    context "when there are expiring registrations from AD users" do
+      before do
+        create_list(:registration, 3, :ad_registration)
+        allow(WasteExemptionsBackOffice::Application.config).to receive(:ad_letters_exports_expires_in).and_return(35)
+        WasteExemptionsEngine::RegistrationExemption.update_all(expires_on: 35.days.from_now)
+        expect(DefraRuby::Aws).to receive(:get_bucket).and_return(bucket)
```
Cruikshanks, line 18: Is there a difference between `allow` and `expect` in this scenario? If not I'd look to use `allow` for all of them.

spec/services/ad_renewal_letters_export_service_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      it "load a file to a AWS bucket and record the content created" do
+        expect(Airbrake).to_not receive(:notify)
+
+        expect { described_class.run }.to change { WasteExemptionsEngine::AdRenewalLettersExport.count }.by(1)
+      end
+
+      context "when one registration is in a broken status and a PDF cannot be generated for it" do
+        it "raises an error on Airbrake but continue the generation for the other letters" do
```
Cruikshanks: I'm not sure I am seeing how this proves if one registration errors it won't stop the export from continuing. Would checking that a `AdRenewalLettersExport` is still created and properly updated be a way of doing it?

It might just be I missed something.

app/services/ad_renewal_letters_export_service.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def ad_letters_expires_on
+    WasteExemptionsBackOffice::Application.config.ad_letters_exports_expires_in.days.from_now.to_date
+  end
+
+  def record_content_created
+    WasteExemptionsEngine::AdRenewalLettersExport.create!(
+      expires_on: ad_letters_expires_on,
+      number_of_letters: ad_expiring_registrations.count,
```
Cruikshanks: I don't think we can use this. When I saw this I checked with @glancyea what the number of letters value will represent in the UI, and we agreed it should be the number of letters generated.

This will return the number of registrations eligible for sending a renewal letter, so we'll somehow have to track how many _actually_ got generated and use that value here instead.

**Update** Displaying an extra message for PDF's that had failed letters is going to be handled as a separate story. However I still think the number of letters represents the number actually in the PDF, not how many we _tried_ to generate a letter for. Hence I'm leaving this comment. Hope that makes sense.

app/views/renewal_letter/bulk.html.erb
```diff
@@ -0,0 +1,25 @@
...
+</head>
+
+<body>
+  <% presenters.each do |presenter| %>
+    <!-- We don't want the bulk to fail if a single render fails -->
+    <%=
+      begin
+        render("renewal_letter/renewal_letter_content", presenter: presenter)
+      rescue StandardError => e
+        Airbrake.notify e, registration_reference: presenter.reference
```
Cruikshanks, line 17: ❤️ Thank you for adding the reference in the meta data!

spec/services/ad_renewal_letters_export_service_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+          registration.registration_exemptions.update_all(expires_on: 35.days.from_now)
+
+          expect(Airbrake).to receive(:notify)
+
+          described_class.run
+        end
+      end
+    end
+
+    context "when an error happen" do
```
Cruikshanks:
```suggestion
    context "when an error happens" do
```

spec/services/ad_renewal_letters_export_service_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        expect(bucket).to receive(:load).and_return(result)
+      end
+
+      it "load a file to a AWS bucket and record the content created" do
+        expect(Airbrake).to_not receive(:notify)
+
+        expect { described_class.run }.to change { WasteExemptionsEngine::AdRenewalLettersExport.count }.by(1)
+      end
+
+      context "when one registration is in a broken status and a PDF cannot be generated for it" do
```
Cruikshanks:
```suggestion
      context "when one registration is in an invalid state and a PDF cannot be generated for it" do
```

spec/services/ad_renewal_letters_export_service_spec.rb
Cruikshanks:
```suggestion
        it "raises an error on Airbrake but continues generation for the other letters" do
```

spec/services/renewal_letters_bulk_pdf_service_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+RSpec.describe RenewalLettersBulkPdfService do
+  describe ".run" do
+    let(:registrations) { create_list(:registration, 2) }
+    let(:result) { double(:result) }
+
+    it "returns a PDF string with a letter per-registration" do
+      expect(Airbrake).to_not receive(:notify)
+
+      expect(described_class.run(registrations)).to start_with("%PDF")
```
Cruikshanks, line 13: This is the same expectation for the test `when there are no registrations`, which for me feels like we're not really testing the 2 contexts.

It'd be great if there was something else we could check that confirms there is a letter per registration, else I'd just update the test statement

```ruby
      it "returns a PDF" do
```

Cruikshanks (changes requested):
There is a mix of questions and requests in here, plus some grammar tweaks. Ping me if you need anything else or want to cover anything further.

spec/services/ad_renewal_letters_export_service_spec.rb
cintamani: Oh no. This allow should actually become an `expect().twice`.

spec/services/ad_renewal_letters_export_service_spec.rb
cintamani: Well, the complication of doing that is that we need the ability to read PDF content which we currently don't have. Since the test is the same as the above one (same setup) and I don't `raise` any other error as I test it outside of this context, then I am 99% sure the error occuring is coming from the extra `registration` I am creating in this scenario which is in a broken status (no addresses). Open on suggestions on how to make it clearer :)

app/services/ad_renewal_letters_export_service.rb
cintamani: Yes, I did though about this scenario. The problem is that we can't generate the letters separately outside of the view with the current code. I was thinking one way to make this check would be to check that the presenter can respond without error to every public interface method once initialized with a registration, and when it cannot then we decide that is a failed generation. Still to decide is how we record this "missing/broken" letter in the DB so that we can show the info to the user. 
Anyway, I don't find this scenario important enough and I think we can tackle this problem separately and even release it later on

spec/services/ad_renewal_letters_export_service_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+RSpec.describe AdRenewalLettersExportService do
+  describe ".run" do
+    let(:bucket) { double(:bucket) }
+    let(:result) { double(:result, successful?: true) }
+
+    context "when there are expiring registrations from AD users" do
+      before do
+        create_list(:registration, 3, :ad_registration)
+        allow(WasteExemptionsBackOffice::Application.config).to receive(:ad_letters_exports_expires_in).and_return(35)
```
cintamani:
```suggestion
        expect(WasteExemptionsBackOffice::Application.config).to receive(:ad_letters_exports_expires_in).and_return(35).twice
```

spec/services/renewal_letters_bulk_pdf_service_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe RenewalLettersBulkPdfService do
+  describe ".run" do
+    let(:registrations) { create_list(:registration, 2) }
+    let(:result) { double(:result) }
+
+    it "returns a PDF string with a letter per-registration" do
```
cintamani:
```suggestion
    it "returns a PDF string" do
```

spec/services/renewal_letters_bulk_pdf_service_spec.rb
cintamani: We can't do that without adding the ability to read PDF content 🤔 shall I research that? I can probably bring in Prawn in the `:test` group only.

spec/services/ad_renewal_letters_export_service_spec.rb
Cruikshanks: I think the issue with just checking Airbrake gets notified is that it would happen if one letter fails, or if the whole job failed.

However if the whole job were to fail, I'm assuming the `AdRenewalLettersExport` record would not get created. Whereas if we got the notification because just one letter failed, then the record would still be created.

Tie this into only recording the number of letters that were exported successfully, we could possibly add this to the test

```ruby
expect { described_class.run }.to change { WasteExemptionsEngine::AdRenewalLettersExport.count }.by(1)
expect(WasteExemptionsEngine::AdRenewalLettersExport.last.number_of_letters).to eq(3)
```

app/services/ad_renewal_letters_export_service.rb
Cruikshanks: Then I think we need to drop handling this scenario at the moment. The number of letters needs to represent the number successfully generated and included in the PDF, not just the number expiring.

So if we are not ready to handle recording partial successes, then we need to stop and cancel the whole job and record it as failed when one blows up. We can then look to bring back in partial generated letters as part of the new story we've raised.

spec/services/renewal_letters_bulk_pdf_service_spec.rb
Cruikshanks: I'd lean towards

- just checking we have a PDF for the first test (there are registrations)
- changing the design and returning `nil` (when there are no registrations)

On that last point it kind of makes more sense to me. If there are no registrations I would not expect to generate a PDF, rather than generate a blank/empty one.

spec/services/ad_renewal_letters_export_service_spec.rb
cintamani: Sounds good :)

app/services/ad_renewal_letters_export_service.rb
cintamani: Oh. This makes me feel bad really :/ 
So, 2 questions:
 - Why is it so important that the number there is exactly the number of Letters in the PDF?  
 - After looking at the wireframe, what is the whole point of showing the user a `Partial fail` without telling them which letters have failed? Do we really need that?

Cruikshanks: Started writing some glorious reasons as to why we've got this decision, only to then...

pivot

The thing I've come to see is if we leave the number as what ***should*** have been exported, NCCC will then be able to check for themselves based on the generated PDF whether it matches up.

Ideally the service should be able to tell them something didn't export, rather than having to manually tally the PDF to the figure on screen. However we're all agreed that will come as a later story.

So the main thing I've pivoted on is that there is more value in having **Number of letters** represent how many were eligible, instead of just how many were printed (as we can determine that number from the PDF).

spec/services/ad_renewal_letters_export_service_spec.rb
Cruikshanks: You will have already probably spotted this, but as we're no longer capturing the number that got printed, there is no point in having the second line in my example. Just checking a record got created would be enough.

Cruikshanks (approved):
shipit

PR #386, opened 2019-09-12
Ad renewal letters - Create record first

cintamani:
This refactoring includes:
 - Rake task to use a `letters` namespace so that we can reuse it for other letters exports
 - We decided to create the export letters record first and update it with failure or success info later. This will make it easier to create a button in the backend that can re-run the job based on the record.
 - Everything run around the expires_on date: that is what determine the correct content and date of a given bulk document


Cruikshanks (approved):

lib/tasks/letters.rake
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+namespace :letters do
+  namespace :export do
```
Cruikshanks, line 4: Could we tweak the namespace name to be `:generate`? This will then make it consistent with the reports rake task.

Cruikshanks (changes requested):

lib/tasks/letters.rake
cintamani: As discussed in slack, I'll open a separate PR for renaming the `reports` stuff instead to use `export` and renaming the `reports` scope to `data`

Cruikshanks (approved):
🍾 🎉

PR #389, opened 2019-09-13
Add dashboard view to AD letters

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-582

Create a new letters dashboard linked in the menu for agents to use.
The dashboard show information about AD renewal letter exports and gives users the ability to mark an export as printed.

Screenshot 2019-09-13 at 14 09 52


app/assets/stylesheets/partials/_people.scss
```diff
@@ -4,4 +4,5 @@ input.button-link {
   color: $link-colour;
```
Cruikshanks, line 4: This file is duplicated in the front office as well. And with this change it means the code and the behaviour will now be out of sync between the front and back office.

Would it be possible to refactor this file to be in the engine, so we have it in just one place? (I can't remember if having assets in the engine works or not).

Also as it's now not just used by the people page, I think we should rename it something more meaningful, like `_button_link.scss`.

Appreciate this situation has nothing to do with you. More a case of "you touched it last!"

What do you think?

app/assets/stylesheets/partials/_tags.scss
```diff
@@ -1,6 +1,6 @@
 // STATUS TAGS for dashboard and registration details page
 // Taken from https://dwp-design-examples.herokuapp.com/example/tags
-span.status-tag {
+@mixin status-tag-mixin {
```
Cruikshanks, line 3: Just curious. What does this change do?

app/views/ad_renewal_letters_exports/index.html.erb
```diff
@@ -0,0 +1,60 @@
+<div class="grid-row">
```
Cruikshanks, line 1: Do you think there would be any value to creating a `LettersPresenter`? I'm wondering if it would

- allow us to move some of the complexity out of the view into a class, where it might be more clearer
- give us something to test, particularly some of the more nested if-else-if conditions in here

Also trying not to look too far ahead to a confirmation letters export, but I'm betting we'd get some reuse out of it when we came to implement that feature.

config/initializers/time_formats.rb
```diff
@@ -2,6 +2,8 @@
 
 Time::DATE_FORMATS[:time_on_day_month_year] = "%-l:%M%P on %-d %B %Y"
 Date::DATE_FORMATS[:month_year] = "%B %Y"
+Date::DATE_FORMATS[:abbr_day_month] = "%-d %b"
```
Cruikshanks, line 5: ❤️ Well remembered we have these!

spec/helpers/ad_renewal_letters_exports_helper_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe AdRenewalLettersExportsHelper, type: :helper do
+  describe "#format_email_to_name" do
+    it "takes an email and format it into a new string" do
```
Cruikshanks:
```suggestion
    it "takes an email and returns a string formatted as 'Firstname Lastname'" do
```

spec/helpers/ad_renewal_letters_exports_helper_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+      expect(helper.format_email_to_name("super_agent@wex.gov.uk")).to eq("Super Agent")
```
Cruikshanks: Could we change the example to use a period?

```suggestion
      expect(helper.format_email_to_name("katherine.johnson@nasa.org.uk")).to eq("Katherine Johnson")
```

We seed our test data with names that use underscores, but in the actual app 99% of the time the time it'll be a period.

Also is it worth having a scenario that covers emails without any separator, just to ensure we don't get an exception if we encounter one?

N.B. See comment on the helper itself, as I'm not sure it will work if the email uses a `.`.

app/helpers/ad_renewal_letters_exports_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module AdRenewalLettersExportsHelper
+  def format_email_to_name(email)
+    email.scan(/(\A.*)_(.*)@/).flatten.map(&:capitalize).join(" ")
```
Cruikshanks: My knowledge of regex is shocking. So I used rubular to do a quick test https://rubular.com/r/EGxeUwzeNM1Qew

Based on that I think this will work fine for our seeded test email accounts, but will fail in production where they will be separated with a `.`.

spec/models/waste_exemptions_engine/ad_renewal_letters_export_spec.rb
```diff
@@ -21,5 +21,41 @@ module WasteExemptionsEngine
...
+        end
+      end
+
+      context "when printed_on is empty" do
+        it "is not printed" do
+          expect(ad_renewal_letters_export).to_not be_printed
+        end
+      end
+
+      context "when prited_by is empty" do
```
Cruikshanks:
```suggestion
      context "when printed_by is empty" do
```

app/controllers/ad_renewal_letters_exports_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class AdRenewalLettersExportsController < ApplicationController
+  def index
+    authorize! :read, WasteExemptionsEngine::AdRenewalLettersExport
+
+    @ad_renewal_letters_exports = WasteExemptionsEngine::AdRenewalLettersExport.all
+  end
+
+  def update
+    @ad_renewal_letters_exports = WasteExemptionsEngine::AdRenewalLettersExport.find(params[:id])
```
Cruikshanks: Is the reason `@ad_renewal_letters_exports` is plural in `update()` so that it works with the strong params / `ad_renewal_letters_exports_attributes` function?

If it isn't, I'd just ask if we could make it `@ad_renewal_letters_export` in as we're only dealing with the one instance here.

If it is, don't bother about it. This is just a NIT 😁

Cruikshanks (changes requested):
I think there are going to need to be some changes unless I'm wrong about the regex in the helper!

I've also asked if it would be worth doing some other refactorings, and the odd question.

See you Monday if you want to go through them face-to-face @cintamani 👋

app/assets/stylesheets/partials/_tags.scss
cintamani: This makes it possible to override some rules there, something that you can't do with a simple `  @extend` :)

app/controllers/ad_renewal_letters_exports_controller.rb
cintamani: Oh!! This is sop true! I have added too many `s`s XD

app/helpers/ad_renewal_letters_exports_helper.rb
cintamani: I wasn't sure about the format of the production ones, so I was expecting this :D will update the regex :)

app/views/ad_renewal_letters_exports/index.html.erb
cintamani: I have thought about that, but there are only a couple of methods I would implement there, which are the formatted info, and I am not sure is worth creating a presenter for them. 
I do find all the logic here quite confusing myself, but the business rules are complex themselves, and I can't see a better way of getting rid of all these `if` and `else` and `elseif`. 
A presenter cannot deal with HTML, so probably we should do a better use of the helper instead? Would be good to pair on this on Monday and try few things until we are happy with it :)

app/helpers/ad_renewal_letters_exports_helper.rb
cintamani:
```suggestion
    email.scan(/(\A.*)\.(.*)@/).flatten.map(&:capitalize).join(" ")
```

Cruikshanks (approved):
❤️ 🎉 🍾 🎈 🎂

PR #390, opened 2019-09-13
Schedule export AD renewal letters

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-581


.env.example
```diff
@@ -37,6 +37,7 @@ EXPIRE_REGISTRATION_EXEMPTION_RUN_TIME="00:05"
 # Exporter config
 EXPORT_SERVICE_BATCH_SIZE=10
 EXPORT_SERVICE_EPR_EXPORT_TIME="1:05"
+EXPORT_SERVICE_AD_RENEWAL_LETTERS_TIME="2:05"
```
cintamani:
```suggestion
EXPORT_SERVICE_AD_RENEWAL_LETTERS_TIME="1:05"
```

.travis.yml
```diff
@@ -21,6 +21,7 @@ env:
     - EXPORT_SERVICE_BULK_EXPORT_FREQUENCY="friday"
     - EXPORT_SERVICE_BULK_EXPORT_TIME="16:05"
     - EXPORT_SERVICE_BOXI_EXPORT_TIME="03:05"
+    - EXPORT_SERVICE_AD_RENEWAL_LETTERS_TIME="2:05"
```
cintamani:
```suggestion
    - EXPORT_SERVICE_AD_RENEWAL_LETTERS_TIME="1:05"
```

config/schedule.rb
```diff
@@ -50,6 +50,13 @@
   rake "reports:export:boxi"
 end
 
+# This is the daily AD renewal letters export service.
+# Will run once a day in the early morning hours and generate a zip file containing
+# data required for boxi.
+every :day, at: (ENV["EXPORT_SERVICE_AD_RENEWAL_LETTERS_TIME"] || "02:05"), roles: [:db] do
```
cintamani:
```suggestion
every :day, at: (ENV["EXPORT_SERVICE_AD_RENEWAL_LETTERS_TIME"] || "01:05"), roles: [:db] do
```

.env.example
```diff
@@ -37,6 +37,7 @@ EXPIRE_REGISTRATION_EXEMPTION_RUN_TIME="00:05"
 # Exporter config
 EXPORT_SERVICE_BATCH_SIZE=10
 EXPORT_SERVICE_EPR_EXPORT_TIME="1:05"
+EXPORT_SERVICE_AD_RENEWAL_LETTERS_TIME="1:05"
```
irisfaraway, line 40: Is it OK to have this scheduled at the same time as the EPR export? Feels like they shouldn't be clashing.

config/schedule.rb
```diff
@@ -50,6 +50,13 @@
   rake "reports:export:boxi"
 end
 
+# This is the daily AD renewal letters export service.
+# Will run once a day in the early morning hours and generate a zip file containing
+# data required for boxi.
```
irisfaraway:
```suggestion
# all AD renewal letters expiring in X days.
```

spec/schedule_spec.rb
```diff
@@ -63,6 +63,13 @@
     expect(job_details[:every][1][:at]).to eq(ENV["EXPIRE_REGISTRATION_EXEMPTION_RUN_TIME"])
   end
 
+  it "takes the expire ad renewal letters execution time from the appropriate ENV variable" do
```
irisfaraway:
```suggestion
  it "takes the export AD renewal letters execution time from the appropriate ENV variable" do
```

config/schedule.rb
irisfaraway: The zip file bit may be inaccurate as well – not sure

irisfaraway (changes requested):
A few bits, but nothing major 👍

config/schedule.rb
cintamani: Copy paste fail! D:

.env.example
cintamani: The EPR exports in prod actually run at 22.05 :) We should fix these values to reflect the production once. For another PR :)

config/schedule.rb
```diff
@@ -50,6 +50,13 @@
   rake "reports:export:boxi"
 end
 
+# This is the daily AD renewal letters export service.
+# Will run once a day in the early morning hours and generate a zip file containing
```
cintamani:
```suggestion
# Will run once a day in the early morning hours and generate a PDF file containing
```

irisfaraway (approved):

PR #397, opened 2019-09-17
Cast string to integer

cintamani:


andrewhick (approved):
Looks right to me, but I'll let @Cruikshanks confirm too :)

Cruikshanks (approved):
Happy to approve this, but I would like to see some context in the commit message as to why this change was made.

andrewhick:
Some background:
When I tried to run the rake task to generate the letters, it failed with the following message:
```
NoMethodError: undefined method `days' for "35":String
/vagrant/waste-exemptions-back-office/lib/tasks/letters.rake:7:in `block (3 levels) in <top (required)>'
```
This change converts the environment variable to an integer so that it can be read.

PR #400, opened 2019-09-18
Clean old AD renewal letters files

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-635

This adds a new `deleted` status to the AD renewal letters export records.
The enum-defined `deleted!` method is then overriden to also delete the AWS file from the bucket before updating the status of the records to `deleted`. The dashboards result have been filtered out to not show records in a `deleted` status.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -60,7 +60,8 @@ module WasteExemptionsBackOffice
     config.export_batch_size = ENV["EXPORT_SERVICE_BATCH_SIZE"] || 1000
 
     config.years_before_expiry = ENV["YEARS_BEFORE_EXPIRY"] || 3
-    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 35
+    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 35 # Days
+    config.ad_letters_delete_records_in = ENV["AD_LETTERS_DELETE_RECORDS_IN"] || 3 # Weeks
 
     # Emails
     config.email_test_address = ENV["EMAIL_TEST_ADDRESS"]
--- a/config/application.rb
+++ b/config/application.rb
@@ -60,8 +60,8 @@ module WasteExemptionsBackOffice
     config.export_batch_size = ENV["EXPORT_SERVICE_BATCH_SIZE"] || 1000
 
     config.years_before_expiry = ENV["YEARS_BEFORE_EXPIRY"] || 3
-    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 35 # Days
-    config.ad_letters_delete_records_in = ENV["AD_LETTERS_DELETE_RECORDS_IN"] || 3 # Weeks
+    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 35
+    config.ad_letters_delete_records_in = ENV["AD_LETTERS_DELETE_RECORDS_IN"] || 21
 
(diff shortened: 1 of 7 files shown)


config/application.rb
```diff
@@ -60,7 +60,8 @@ class Application < Rails::Application
     config.export_batch_size = ENV["EXPORT_SERVICE_BATCH_SIZE"] || 1000
 
     config.years_before_expiry = ENV["YEARS_BEFORE_EXPIRY"] || 3
-    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 35
+    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 35 # Days
+    config.ad_letters_delete_records_in = ENV["AD_LETTERS_DELETE_RECORDS_IN"] || 3 # Weeks
```
Cruikshanks: Can we make this `days`? I understand that for production we have been specifying 3 weeks, but we can just as easily represent that as `21 # Days`.

My reasoning is to give us flexibility in non-production environments for testing.

lib/tasks/letters.rake
```diff
@@ -10,6 +10,10 @@ namespace :letters do
         expires_on: expires_on
       ).export!
 
+      older_than = WasteExemptionsBackOffice::Application.config.ad_letters_delete_records_in.to_i.weeks.from_now
```
Cruikshanks: See comment on change in `application.rb` regards using `days` instead of `weeks`.

spec/services/ad_renewal_letters_export_cleaner_service_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+require "rails_helper"
+
+RSpec.describe AdRenewalLettersExportCleanerService do
+  describe ".run" do
+    let(:bucket) { double(:bucket) }
+    let(:ad_renewal_letters_export) do
+      create(:ad_renewal_letters_export, created_at: 3.weeks.ago, file_name: "example.txt")
+    end
+
+    it "mark as deleted all records oldert than a specified date and delete their associated files" do
```
Cruikshanks:
```suggestion
    it "marks as deleted all records older than a specified date and deletes their associated files" do
```

Cruikshanks (changes requested):
My main **Request changes** is to use days rather than weeks to define the time period for when letters get deleted.

config/application.rb
cintamani:
```suggestion
    config.ad_letters_delete_records_in = ENV["AD_LETTERS_DELETE_RECORDS_IN"] || 21
```

config/application.rb
```diff
@@ -60,7 +60,8 @@ class Application < Rails::Application
     config.export_batch_size = ENV["EXPORT_SERVICE_BATCH_SIZE"] || 1000
 
     config.years_before_expiry = ENV["YEARS_BEFORE_EXPIRY"] || 3
-    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 35
+    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 35 # Days
```
cintamani:
```suggestion
    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 35
```

lib/tasks/letters.rake
cintamani:
```suggestion
      older_than = WasteExemptionsBackOffice::Application.config.ad_letters_delete_records_in.to_i.days.from_now
```

Cruikshanks (approved):

PR #403, opened 2019-09-18
Switch source of defra_ruby_aws to rubygems

Cruikshanks:
Prior to this change the source for the [defra_ruby_aws](https://github.com/DEFRA/defra-ruby-aws) gem was GitHub. Essentially any push to master would cause [dependabot](https://dependabot.com/) to raise a PR.

That's fine whilst we're under active development and WEX is the only service using the gem, but we're now about to start a period of focusing on WCR and are likely to introduce the gem there as well.

This might mean further changes are made, but we'll want to have a bit more control over when these get applied back in WEX.

Plus it breaks our currently unwritten convention; our engines are sourced from GitHub, but our gems are sourced from Rubygems.

So this change switches the source for the `defra_ruby_aws` gem to [rubygems](https://rubygems.org/), and uses pessimistic versioning to lock the accepted updates to those that should not break the current app (assuming we stick to using semantic versioning!)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -61,9 +61,7 @@ gem "waste_exemptions_engine",
     branch: "master"
 
 # Use the Defra Ruby Aws gem for loading files to AWS buckets
-gem "defra_ruby_aws",
-    git: "https://github.com/DEFRA/defra-ruby-aws",
-    branch: "master"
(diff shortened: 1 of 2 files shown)


cintamani (approved):
💯

PR #405, opened 2019-09-19
Fix bug on Cleaning AD renewal letters

cintamani:
This fix a bug in the cleaning of AD renewal letters where the date passed to the service was in the future rather than in the past


Cruikshanks (approved):
🤦‍♂ Should have spotted this when I reviewed the previous PR

PR #415, opened 2019-09-23
Add area lookup for addresses task

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-487

Add task and scheduler for the area lookup background job responsible for updating addresses that miss area information.


app/services/ea_area_lookup_service.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class EaAreaLookupService < ::WasteExemptionsEngine::BaseService
```
Cruikshanks: How would you feel about renaming this service? As per other PR's I'd want to drop the EA prefix. It's already very similar to `AreaLookupService` and with that dropped we'll get a clash.

 Plus I think `AreaLookupService` is handling the lookup, this is more about updating the record.

So suggestion time! Mine are

- `UpdateAreaService`
- `MissingAreaService`
- `UpdateMissingAreaService`

N.B. Although the reason we've built it is to update missing areas, if you look at the code it doesn't care whether the area field is populated, so my favourite is `UpdateAreaService`, but it's just an opinion!

Cruikshanks (changes requested):
Just the naming of the service is the reason I've asked for a change. TL;DR I feel its a little too similar to `AreaLookupService` in the engine.

app/services/ea_area_lookup_service.rb
cintamani: Yes, you are right! I think `UpdateAreaService` is the best one. Thanks for the suggestions :)

Cruikshanks (approved):

PR #417, opened 2019-09-24
Add comment in regard of design needed in 2 pages

cintamani:


app/views/renewal_letter/_renewal_letter_content.html.erb
```diff
@@ -1,3 +1,6 @@
+<!-- WARNING: Before changing this design, remember that this have been created by Martin to always take 2 pages and never exceed them.
```
cintamani:
```suggestion
<!-- WARNING: Before changing this design, remember that this has been created by Martin to always take 2 pages and never exceed them.
```

andrewhick (approved):
Thanks for adding - this reduces the risk of us breaking this in future :)

PR #426, opened 2019-09-26
Restrict missing area update job to site addresses

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-487

The aim is to try and identify the EA area for each registration we have.

The address we use to determine this though is the **site**. Where the site is located is what determines a registration's area. So we don't care if the x & y or area is missing on the other addresses (operator and contact). We only care they are populated for the `:site` address.

So this change amends the query in the lookup to only return site addresses where x & y is populated but area is missing.

### DIFF
--- a/lib/tasks/lookups.rake
+++ b/lib/tasks/lookups.rake
@@ -2,11 +2,11 @@
 
 namespace :lookups do
   namespace :update do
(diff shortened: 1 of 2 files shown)


cintamani:
Oh! This make sense!! XD

PR #428, opened 2019-09-26
Remove dependence on env vars for some tests

Cruikshanks:
During the housekeeping of the env vars in PR #427 it was noted that we default most env vars, but because our `schedule_spec.rb` tests specifically for the values of certain env vars we are still forced to set them across all environments.

We've decided that we can be confident calling `ENV["MY_VAR"]` will return the value of `MY_VAR` if set as an environment variable, and therefore the various cron jobs `schedule.rb` manages will have the correct time set.

So this change

- updates the `schedule_spec.rb` tests to just confirm the default values
- removes env vars that are defaulted in the app from `.travis.yml`


Cruikshanks:
I believe the other broken test is as a result of running the test suite on Travis in 'the witching' hour i.e. between 00:00 and 01:00 BST, when Travis is on UTC.

Will re-run the tests tomorrow am to confirm 😁 🤞

Cruikshanks:
GitHub is playing silly b#ggers. I have re-run the build and it passed, just for some reason GitHub is not updating.

PR #429, opened 2019-09-27
Add support for rake integration tests

Cruikshanks:
This change will allow us to invoke rake tasks within out test suite to ensure that they run as expected.

Supporting rake integration tests comes with 2 primary challenges

- ensuring accurate test coverage
- ensuring any records created are correctly cleaned

Both result from the fact that invoking a rake task is the equivalent of starting a new instance. For simplecov this has the effect of it only capturing the coverage for the last rake task invoked. For DatabaseBase cleaner it is not called because the connection to the DB is new and not the one it is applying its transaction strategy to.

Each has its own solution. For simplecov we need to start an instance of Rake and load all tasks before the suite starts. This means its able to distinguish between which files are being called and collate the test coverage, rather than keep overwriting.

For DatabaseCleaner we need to mark our tests, and then use Rspec's config to alter the cleaning strategy when a rake test is called, plus clean the database after each rake spec.

### DIFF
--- a/lib/close_airbrake.rb
+++ b/lib/close_airbrake.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# To ensure errors are reported by Airbrake when a rake task fails we need
+# to tell Airbrake to close before the exit the task. This is because
+# Airbrake operates asynchronously, queuing up the notifications to send rather
+# than forcing the app to wait.
+#
+# However there is an issue when calling `Airbrake.close` in the test
+# environment. The test suite will complain about airbrake being closed already
+# when running. Since there is no way in version 5.8 to ask Airbrake if it is
+# already closed or to reopen it before every test, this check prevents the test
+# suite from complaining.
+#
+# We could have just put the line `Airbrake.close unless Rails.env.test?` in
+# each rake task, but we wanted to record these notes but just in one place.
+class CloseAirbrake
+  def self.now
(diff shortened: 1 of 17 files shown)


cintamani (dismissed):
This looks good! Shall we also update the guides and make sure this tests are always at an `integration` level? We can also think about moving them into `spec/integration`

Cruikshanks:
I agree we should document our approach to rake tasks and tests, but I'm not sure I agree with running them at the integration level.

I'm not planning on getting rid of what I've done (it's taken ages to get it working 😩 !) but now I've spent a bunch of time with it, I've come to the conclusion that if the missing area task had been created as a `MissingAreaService` then we wouldn't have needed to add this support. We could have tested the service class as we do with the others. And I don't believe we would have lost much.

Going forward I'd push for an approach where

- no rake task should contain any logic beyond calling a service class
- the service should of course have unit tests
- the rake task itself is just covered by a basic 'when invoked it doesn't raise an error' test

cintamani:
I think we can do both :) Integration tests at rake task level, unit tests for services

cintamani (approved):

Cruikshanks:
> I think we can do both :) Integration tests at rake task level, unit tests for services

Sorry, that was what I meant! I'd just be aiming for the integration tests to be the basic 'does it run' but anything else to be standard unit tests against a service.

PR #421, opened 2019-09-25
Add lookup job for X and Y updates on addresses

cintamani:
From: RUBY-468

Add rake task to update easting and northing information in all addresses missing it.
The job will use OS places to update addresses that have the `x` and/or `y` attributes set to nil and a postcode.
It will use the OS maps for fetching `x` and `y` from grid references.
The job will run for a maximum amount of time specified in the ENV variable named `EASTING_AND_NORTHING_LOOKUP_RUN_FOR`.
The job is also scheduled to run daily at a time specified by the ENV variable named `EASTING_AND_NORTHING_LOOKUP`.

PR #430, opened 2019-09-27
Update schedule.rb order by runtime

Cruikshanks:
This is just some housekeeping of the `config/schedule.rb` class. We have ordered the schedule by runtime so its clearer to see in what order things run during the night for the WEX service.

We have also updated the comments to be consistent in their explanations, and how we set the time for each one.


Cruikshanks:
This is in draft until #421 is merged, at which point it should become a much smaller PR 😁 !

Cruikshanks:
I wasn't finished GitHub!

cintamani (approved):

PR #436, opened 2019-10-02
Add with_postcode to x & y lookup job scope

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-486

This fixes an issue encountered in testing with certain addresses. Some of the old data we migrated when launching the new version of the service was missing site addresses. So the migrating created blank site records in these cases as the new code expects there to be 1 site address for every registration.

However these have neither an x & y nor a postcode. We had not coded for this in our solution, so the records were coming up in our query, and we were trying to look up the x & y with OS Places using a null postcode.

We added the `with_postcode` scope to the engine. This change applies it to our x & y lookup job.


andrewhick (approved):
Thanks Alan, this all makes sense to my brain :shipit:

PR #455, opened 2019-10-25
Exclude NCCC postcodes from automated renewal process

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-720

Registrations with a site postcode of 'S9 4WF' (in any format) should not be emailed about renewals.

---

Note that in a Friday Fail I accidentally merged these to master originally and then reverted them for some immediate damage control. We may need to have a look at tidying up the history later. Sorry!


app/models/waste_exemptions_engine/registration.rb
```diff
@@ -16,6 +18,12 @@ class Registration < ActiveRecord::Base
 
     scope :renewals, -> { where.not(referring_registration_id: nil) }
 
+    scope :contact_email_is_not_nccc, -> { where.not(contact_email: NCCC_EMAIL) }
+
+    scope :site_address_is_not_nccc, lambda {
+      joins(:addresses).merge(Address.site.not_nccc)
```
cintamani, line 24: Pretty coding 😍

cintamani (approved):

PR #487, opened 2020-01-28
Update VCR cassettes

Cruikshanks:
Every couple of weeks we update our VCR cassettes to ensure our mock requests match the current real world ones.


Cruikshanks:
Codeclimate is just being it's usual self. Should wake up once this is merged.

Gemfile.lock
```diff
@@ -157,7 +157,7 @@ GEM
       govuk_frontend_toolkit (>= 6.0.2)
       rails (>= 4.1.0)
       sass (>= 3.2.0)
-    govuk_frontend_toolkit (8.1.0)
+    govuk_frontend_toolkit (9.0.0)
```
cintamani: I am a bit concerned about this update, are we sure we want to apply this without a ticket to QA visual regressions?

cintamani (approved):

PR #573, opened 2020-06-29
Integrate defra-ruby-features engine

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-1139

This integrate the defra-ruby-features engine in the back office app.

Note: There is currently no user with permission to use the feature, due to the fact that there is no existing developer level role, hence it will need to be discussed with the team first


cintamani:
<3

irisfaraway (approved):
👏

PR #590, opened 2020-07-29
Add developer role to back-office

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-1177

This is related to [PR #573 integrating feature toggle support](https://github.com/DEFRA/waste-exemptions-back-office/pull/573).

Our feature toggles integration allows a user in the back-office to 'toggle' features on an off. These are useful when introducing new functionality because it allows us to enable, and disable the feature quickly. This avoids the delays we experience using the RfC process just to toggle env vars.

However, these feature toggles are something we only want the delivery team to control. We'll have a better understanding of the implications of what happens when they are switched on and off, so we need to control them. In [waste-carriers](https://github.com/DEFRA/waste-carriers-back-office) we introduced a developer role and then only authorised it to access the feature toggles page.

We intend to do the same for WEX. So this change covers adding a new 'developer' role to the service.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -16,6 +16,7 @@ class Ability
     permissions_for_super_agent if user.role_is?(:super_agent)
     permissions_for_admin_agent if user.role_is?(:admin_agent)
     permissions_for_data_agent if user.role_is?(:data_agent)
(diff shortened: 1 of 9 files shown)


irisfaraway (approved):
ballmer

PR #591, opened 2020-07-30
Refactor how can? is stubbed in some tests

Cruikshanks:
Whilst working on [PR #573](https://github.com/DEFRA/waste-exemptions-back-office/pull/573) to integrate our new feature toggles engine we exposed an issue with how `can?` has been stubbed in the tests.

The PR adds a new call to `can?` in the apps main layout view `can?(:manage, WasteExemptionsEngine::FeatureToggle, current_user)`. This is to check whether the current user is permitted to access the feature toggles page.

This was no issue in the [waste-carriers-back-office](https://github.com/DEFRA/waste-carriers-back-office) because there `can?()` is stubbed local to each test using standard rspec stubs.

In WEX though the `spec/helpers/action_links_helper_spec.rb` includes a module only loaded when the tests are run that adds a custom `can?()` to `ActionLinksHelper`.

```ruby
module ActionLinksHelper
  def can?(_action, _resource)
    false
  end
end
```

This is hardcoded to except only 2 arguments. Our addition of the new check-in the layout means every request test that passes through the dashboard (so all of them!) now fails with

```
 ActionView::Template::Error:
   wrong number of arguments (given 3, expected 2)
 # ./spec/helpers/action_links_helper_spec.rb:7:in `can?'
 # ./app/views/layouts/application.html.erb:40:in `block in _app_views_layouts_application_html_erb___2225752769677955993_109860'
 # ./app/views/layouts/application.html.erb:10:in `_app_views_layouts_application_html_erb___2225752769677955993_109860'
 # ./spec/requests/user_activations_spec.rb:56:in `block (5 levels) in <top (required)>'
 # ./spec/support/database_cleaner.rb:16:in `block (3 levels) in <top (required)>'
 # ./spec/support/database_cleaner.rb:15:in `block (2 levels) in <top (required)>'
 # ------------------
 # --- Caused by: ---
 # ArgumentError:
 #   wrong number of arguments (given 3, expected 2)
 #   ./spec/helpers/action_links_helper_spec.rb:7:in `can?'
```

I'm sure we could try and work around the need to call `can?()` with 3 args in the layout and everything would work again.

But then we are adding to the inconsistencies between the 2 services rather than removing them. So this PR adds some grease to our elbow and gets stuck in updating the WEX stubbing of `can?()` to match WCR!

we_can_do_it

### DIFF
--- a/spec/helpers/action_links_helper_spec.rb
+++ b/spec/helpers/action_links_helper_spec.rb
@@ -2,13 +2,6 @@
 
 require "rails_helper"
 
-# Add #can? to the module so it can be stubbed in the tests
-module ActionLinksHelper
-  def can?(_action, _resource)
-    false
-  end
-end
-
 RSpec.describe ActionLinksHelper, type: :helper do
   include Devise::Test::ControllerHelpers
 
@@ -110,8 +103,10 @@ RSpec.describe ActionLinksHelper, type: :helper do
     context "when the resource is a registration" do
       let(:resource) { create(:registration) }
 
+      before(:each) { allow(helper).to receive(:can?).with(:update, resource).and_return(can) }
+
       context "when the user has permission to update a registration" do
-        before(:each) { allow_any_instance_of(described_class).to receive(:can?).with(:update, resource).and_return(true) }
+        let(:can) { true }
 
         it "returns true" do
           expect(helper.display_edit_link_for?(resource)).to eq(true)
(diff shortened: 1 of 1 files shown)


Cruikshanks:
Ok, so I thought this was gonna involve going through a bunch-o-tests. Turns out this is all that is needed 😳 

doh

Cruikshanks:
I've made the same change in #573 and it gets everything passing there as well. So just need this merged and then feature toggles is done (once I rebase it 😁 )

irisfaraway (approved):

PR #597, opened 2020-08-28
Set up Notify service for first reminder email

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1184

The second template is extremely similar so doing a base service in preparation.


irisfaraway:
We'll be increasing test coverage when we add the second reminder email, so happy to let SonarCloud go for now.

PR #698, opened 2021-04-08
Resend confirmation email on demand

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1359

We want to allow back office users who are already able to send renewal emails, renewal letters and confirmation letters to also re-send confirmation emails when needed. These are sent to the contact and applicant email.


app/controllers/resend_confirmation_email_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    redirect_back(fallback_location: root_path)
+  end
+
+  private
+
+  def registration
+    @_registration ||= WasteExemptionsEngine::Registration
+                       .includes(registration_exemptions: :exemption)
+                       .find_by(reference: params[:reference])
```
tobyprivett, line 28: I wonder whether the reference param could be assigned in the `new` method, or perhaps as a `before_action`. I don't think it matters either way - I just didn't know it was there until I saw the route.

app/controllers/resend_confirmation_email_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      flash_success(success_message)
+    rescue StandardError => e
+      Airbrake.notify e, registration: registration.reference
+      Rails.logger.error "Failed to send confirmation email for registration #{registration.reference}"
+
+      flash_error(failure_message, failure_description)
+    end
+
+    redirect_back(fallback_location: root_path)
```
tobyprivett, line 20: I hadn't seen `fallback_location` before - nifty!

spec/requests/resend_confirmation_email_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+
+    context "when an admin agent user is signed in", disable_bullet: true do
+      let(:user) { create(:user, :admin_agent) }
+
+      around do |example|
+        # Bullet can't make up its mind about whether or not we should eager-load
+        # `registration_exemptions: :exemption` - if we have it, it tells us to
+        # remove it, and if we don't have it, it tells us to add it.
+        # In this case we're leaving it in and telling Bullet to pipe down.
+        Bullet.unused_eager_loading_enable = false
```
tobyprivett, line 32: That's annoying! I spent a few minutes looking into this and didn't find a solution.

tobyprivett (approved):
A pleasure to read clean and crisp code! 
I probably need to spend more time with the code in general to get familiar with the design patterns. From what I have seen, this looks to be consistent with how things are currently done.

The controller has a fair bit of business logic and my first instinct would be to extract into a service object. But: this is a probably a single responsibility class and the standards docs advise against chucking everything into a service object.

Nice work!

PR #702, opened 2021-04-26
Remove config for AWS letters bucket

irisfaraway:
We no longer use this bucket as all letters go through Notify.

### DIFF
--- a/config/initializers/defra_ruby_aws.rb
+++ b/config/initializers/defra_ruby_aws.rb
@@ -34,16 +34,6 @@ DefraRuby::Aws.configure do |c|
     encrypt_with_kms: ENV["AWS_BOXI_ENCRYPT_WITH_KMS"]
   }
 
-  letters_export_bucket = {
-    name: ENV["AWS_LETTERS_EXPORT_BUCKET"],
-    region: ENV["AWS_REGION"],
-    credentials: {
-      access_key_id: ENV["AWS_LETTERS_EXPORT_ACCESS_KEY_ID"],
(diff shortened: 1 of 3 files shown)


config/initializers/defra_ruby_aws.rb
```diff
@@ -34,16 +34,6 @@
...
-    region: ENV["AWS_REGION"],
-    credentials: {
-      access_key_id: ENV["AWS_LETTERS_EXPORT_ACCESS_KEY_ID"],
-      secret_access_key: ENV["AWS_LETTERS_EXPORT_SECRET_ACCESS_KEY"]
-    },
-    encrypt_with_kms: ENV["AWS_LETTERS_ENCRYPT_WITH_KMS"]
-  }
-
-  c.buckets = [bulk_bucket, epr_bucket, boxi_export_bucket, letters_export_bucket]
+  c.buckets = [bulk_bucket, epr_bucket, boxi_export_bucket]
```
tobyprivett, line 37: `letters_export_bucket` is no longer in the list. Is it needed?

tobyprivett (approved):

config/initializers/defra_ruby_aws.rb
tobyprivett: Aaah: The answer is in the PR title!

irisfaraway: Yep, this was all used for the old version of the letters, which used to be exported in bulk to PDFs overnight. But we don't do that any more so we can get rid of it!

PR #748, opened 2021-10-21
Navigation header edits

tobyprivett:
Remove erroneous link from header (a bug was reported for this: https://eaflood.atlassian.net/browse/RUBY-1642?focusedCommentId=348427)

~~Enable `link_to_unless_current` for navigation links~~

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -19,27 +19,27 @@
     <nav id="proposition-menu">
       <ul id="proposition-links" class="govuk-header__navigation" aria-label="Top Level Navigation">
         <li class="govuk-header__navigation-item">
-          <%= link_to t("layouts.application.menu.dashboard"),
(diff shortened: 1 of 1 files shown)


irisfaraway:
@tobyprivett Where's the warning about redundant links? Just curious why we're changing this as it takes us away from the standard behaviour of the back office apps

tobyprivett:
> @tobyprivett Where's the warning about redundant links? Just curious why we're changing this as it takes us away from the standard behaviour of the back office apps

Hi @irisfaraway . It came up as an 'alert' in the Wave accessibility checker. I spotted the issue with the erroneous link and thought I would squash this warning too. It's not an error and we don't have to implement it.

irisfaraway:
@tobyprivett Aha, interesting. I think I'd rather leave it as-is and keep the services consistent (also maybe it's just personal preference but I find it frustrating when a site's navigation changes around!)

irisfaraway (approved):

PR #845, opened 2022-05-23
[1868] WEX - delete registrations that have been expired/ceased/revoked 7 years prior

d-a-v-e:
https://eaflood.atlassian.net/jira/software/projects/RUBY/boards/374?selectedIssue=RUBY-1868

### DIFF
--- a/app/services/expired_registration_cleanup_service.rb
+++ b/app/services/expired_registration_cleanup_service.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class OldRegistrationCleanupService < ::WasteExemptionsEngine::BaseService
+class ExpiredRegistrationCleanupService < ::WasteExemptionsEngine::BaseService
   def run
     return unless registration_ids.any?
 
@@ -17,14 +17,23 @@ class OldRegistrationCleanupService < ::WasteExemptionsEngine::BaseService
 
   private
 
+  # registration exemptions that were deregistered (ceased or revoked)
+  # or expired over 7 years ago
   def registration_ids
-    @registration_ids ||= WasteExemptionsEngine::Registration
-                          .where("created_at < ?", date)
-                          .pluck(:id)
+    @registration_ids ||=
+      WasteExemptionsEngine::RegistrationExemption
+      .where(
+        <<~SQL
+          deregistered_at <= '#{date}'
+          OR (state = 'expired' AND updated_at <= '#{date}')
+        SQL
+      )
+      .pluck(:registration_id)
+      .uniq
   end
 
-  def registration_ids_sql
-    @registration_ids_sql ||= "(#{registration_ids.join(', ')})"
+  def date
+    @date ||= 7.years.ago.to_date
   end
 
   def clear_registrations!
@@ -56,7 +65,7 @@ class OldRegistrationCleanupService < ::WasteExemptionsEngine::BaseService
     )
   end
 
-  def date
-    @date ||= 7.years.ago
+  def registration_ids_sql
+    @registration_ids_sql ||= "(#{registration_ids.join(', ')})"
   end
 end
--- a/app/services/expired_registration_cleanup_service.rb
+++ b/app/services/expired_registration_cleanup_service.rb
@@ -22,18 +22,19 @@ class ExpiredRegistrationCleanupService < ::WasteExemptionsEngine::BaseService
   def registration_ids
     @registration_ids ||=
       WasteExemptionsEngine::RegistrationExemption
-      .where(
-        <<~SQL
-          deregistered_at <= '#{date}'
-          OR (state = 'expired' AND updated_at <= '#{date}')
-        SQL
+      .where("deregistered_at <= ?", date)
+      .or(
+        WasteExemptionsEngine::RegistrationExemption.expired.where(
+          "updated_at <= ?", date
+        )
       )
       .pluck(:registration_id)
       .uniq
   end
 
+  # returns string in format 2015-05-24
   def date
-    @date ||= 7.years.ago.to_date
+    @date ||= ENV.fetch("EXPIRED_REGISTRATION_CLEANUP_DATE", 7.years.ago.to_date)
   end
 
   def clear_registrations!
@@ -44,28 +45,22 @@ class ExpiredRegistrationCleanupService < ::WasteExemptionsEngine::BaseService
 
   # Versions are auto created by the paper_trail gem
   def clear_versions!
-    ActiveRecord::Base.connection.execute(
-      <<~SQL.squish
-        DELETE FROM versions
-        WHERE item_type = 'WasteExemptionsEngine::Registration'
-        AND item_id IN #{registration_ids_sql}
-      SQL
-    )
+    PaperTrail::Version
+      .where(
+        item_type: "WasteExemptionsEngine::Registration",
+        item_id: registration_ids
+      ).delete_all
   end
 
-  # archive_versions exist because of a schema change
+  # archive_versions exist because of a schema change and has no model backing it.
   # see: https://github.com/DEFRA/waste-exemptions-engine/pull/218
   def clear_versions_archive!
     ActiveRecord::Base.connection.execute(
       <<~SQL.squish
         DELETE FROM version_archives
         WHERE item_type = 'WasteExemptionsEngine::Registration'
-        AND item_id IN #{registration_ids_sql}
+        AND item_id IN "(#{registration_ids.join(', ')})"
       SQL
     )
-  end
-
-  def registration_ids_sql
-    @registration_ids_sql ||= "(#{registration_ids.join(', ')})"
   end
 end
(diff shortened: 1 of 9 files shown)


app/services/expired_registration_cleanup_service.rb
```diff
@@ -0,0 +1,71 @@
...
+          deregistered_at <= '#{date}'
+          OR (state = 'expired' AND updated_at <= '#{date}')
+        SQL
+      )
+      .pluck(:registration_id)
+      .uniq
+  end
+
+  def date
+    @date ||= 7.years.ago.to_date
```
timstone: Can this be a configurable value please? this will allow me to test more easily.

app/services/expired_registration_cleanup_service.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def clear_registrations!
```
timstone, line 62: Can you log the registration number during deleting please? this will allow me to check specifically what registrations have been deleted.

app/services/expired_registration_cleanup_service.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  private
+
+  # registration exemptions that were deregistered (ceased or revoked)
+  # or expired over 7 years ago
+  def registration_ids
+    @registration_ids ||=
+      WasteExemptionsEngine::RegistrationExemption
+      .where(
+        <<~SQL
```
PaulDoyle-EA: Can this be expressed in ActiveRecord syntax rather than raw SQL?

app/services/expired_registration_cleanup_service.rb
```diff
@@ -0,0 +1,71 @@
...
+          deregistered_at <= '#{date}'
+          OR (state = 'expired' AND updated_at <= '#{date}')
+        SQL
+      )
+      .pluck(:registration_id)
+      .uniq
```
PaulDoyle-EA: Might need to also apply a limit to the number returned here, in order to protect service availability - see further comment below.

app/services/expired_registration_cleanup_service.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def date
+    @date ||= 7.years.ago.to_date
+  end
+
+  def clear_registrations!
+    WasteExemptionsEngine::Registration
+      .where(id: registration_ids)
+      .destroy_all
```
PaulDoyle-EA, line 65: This may impact the service if the volume is very large, likely to be the case on the first run. Suggestion: Apply a configurable limit to the select. The cleardown of the large outstanding volume can then be spread over several days in production.

app/services/expired_registration_cleanup_service.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  # Versions are auto created by the paper_trail gem
+  def clear_versions!
+    ActiveRecord::Base.connection.execute(
+      <<~SQL.squish
```
PaulDoyle-EA: As above: ActiveRecord syntax please.

app/services/expired_registration_cleanup_service.rb
```diff
@@ -0,0 +1,71 @@
...
+        AND item_id IN #{registration_ids_sql}
+      SQL
+    )
+  end
+
+  # archive_versions exist because of a schema change
+  # see: https://github.com/DEFRA/waste-exemptions-engine/pull/218
+  def clear_versions_archive!
+    ActiveRecord::Base.connection.execute(
+      <<~SQL.squish
```
PaulDoyle-EA, line 81: As above: ActiveRecord syntax please.

app/services/expired_registration_cleanup_service.rb
```diff
@@ -0,0 +1,71 @@
...
+        DELETE FROM version_archives
+        WHERE item_type = 'WasteExemptionsEngine::Registration'
+        AND item_id IN #{registration_ids_sql}
+      SQL
+    )
+  end
+
+  def registration_ids_sql
```
PaulDoyle-EA: Why `_sql` ? This is just the list of registration IDs, right?

config/schedule.rb
```diff
@@ -42,6 +42,12 @@
   rake "cleanup:transient_registrations"
 end
 
+# This is the old registration cleanup job which will delete all records, paper trails
```
PaulDoyle-EA: (Pedantic point....) "Call it "the cleanup job for old registrations" - the current wording makes it look like it's the job that's old!

PaulDoyle-EA (changes requested):
Consider applying a limit or otherwise limiting the resource impact of the task; also replace raw SQL where possible.

app/services/expired_registration_cleanup_service.rb
PaulDoyle-EA: @timstone just checking, do you mean the regIdentifier as opposed to the database ID?

timstone: Yes the regIdentifier, sorry if that wasn't clear.

app/services/expired_registration_cleanup_service.rb
d-a-v-e: There is no model for the versions_archive so I'll leave this as SQL

app/services/expired_registration_cleanup_service.rb
d-a-v-e: its used in the SQL command further up and formatted as such.

app/services/expired_registration_cleanup_service.rb
d-a-v-e: if it's run daily wouldn't it be the equivalent of a days worth of registrations? (as a days worth of registrations fall into the window of being 7 years old)

d-a-v-e: what kind of number would that be?

app/services/expired_registration_cleanup_service.rb
d-a-v-e: done

app/services/expired_registration_cleanup_service.rb
d-a-v-e: done

app/services/expired_registration_cleanup_service.rb
d-a-v-e: leaving for now

app/services/expired_registration_cleanup_service.rb
d-a-v-e: have removed this method as it's used only once now for `clear_version_archive`

config/schedule.rb
d-a-v-e: done

app/services/expired_registration_cleanup_service.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  private
+
+  # registration exemptions that were deregistered (ceased or revoked)
+  # or expired over 7 years ago
+  def registration_ids
+    @registration_ids ||=
+      WasteExemptionsEngine::RegistrationExemption
+      .where("deregistered_at <= ?", date)
```
PaulDoyle-EA, line 27: `deregistered_at` is of type date, `updated_at` is of type datetime and `date` is of type string. Is this query robust?

db/schema.rb
```diff
@@ -212,6 +212,7 @@
     t.boolean "temp_reuse_applicant_email"
     t.boolean "temp_reuse_operator_address"
     t.string "temp_reuse_address_for_site_location"
+    t.boolean "temp_use_registered_company_details"
```
PaulDoyle-EA: Curious why the schema was updated?

d-a-v-e: not to sure - this happens whenever I `vagrant up` will remove!

app/services/expired_registration_cleanup_service.rb
d-a-v-e: references now go to `Rails.logger.info` 

https://github.com/DEFRA/waste-exemptions-back-office/pull/845/files#diff-3413aedef72933aa883008556057ff4cf68ffbda1ef3d8dd775ff1fd3ba0d548R39-R48

lib/tasks/cleanup.rake
```diff
@@ -6,3 +6,10 @@ namespace :cleanup do
     TransientRegistrationCleanupService.run
   end
 end
+
+namespace :cleanup do
+  desc "Remove expired registrations older than 7 years from the database"
+  task old_regsitrations: :environment do
```
PaulDoyle-EA: Typo: "regsitrations". Maybe call it remove_expired_registrations for clarity?

PaulDoyle-EA (dismissed):

lib/tasks/cleanup.rake
d-a-v-e: thank you - have corrected

PaulDoyle-EA (approved):

PR #852, opened 2022-05-30
[1868] - Use expires_on over updated on

d-a-v-e:


spec/services/expired_registration_cleanup_service_spec.rb
```diff
@@ -4,29 +4,56 @@
...
+          PaperTrail::Version.where(item_type: "WasteExemptionsEngine::Registration").count
+        }.from(paper_trail_count).to(0)
+      end
+    end
+
+    context "when ceased and older than 7 years" do
+      let(:registration) { create(:registration, :with_ceased_exemptions, :with_people) }
+
+      before do
+        registration.registration_exemptions.update_all(deregistered_at: 7.years.ago)
```
PaulDoyle-EA, line 52: Should there also be a test for `deregistered_at` less than 7 years ago not getting deleted?

PR #854, opened 2022-05-31
[1846] Use new template for renewal letter

d-a-v-e:
https://eaflood.atlassian.net/browse/RUBY-1846

### DIFF
--- a/app/presenters/notify_renewal_letter_presenter.rb
+++ b/app/presenters/notify_renewal_letter_presenter.rb
@@ -6,7 +6,11 @@ class NotifyRenewalLetterPresenter < BasePresenter
 
   def expiry_date
     # Currently you can only add exemptions when you register, so we can assume they expire at the same time
-    registration_exemptions.first.expires_on.to_formatted_s(:day_month_year)
+    first_exemption.expires_on.to_formatted_s(:day_month_year)
+  end
+
+  def renewal_window_start_date
+    (first_exemption.expires_on - renewal_window_after_expiry_in_days).to_formatted_s(:day_month_year)
   end
 
   def contact_name
@@ -45,6 +49,10 @@ class NotifyRenewalLetterPresenter < BasePresenter
   private
 
   # Exemptions
+
+  def first_exemption
+    @first_exemption ||= registration_exemptions.first
+  end
 
   def exemption_description(exemption)
     "#{exemption.code} #{exemption.summary}"
@@ -145,6 +153,10 @@ class NotifyRenewalLetterPresenter < BasePresenter
   end
 
   # Utility methods
+
+  def renewal_window_after_expiry_in_days
+    ::WasteExemptionsEngine.configuration.renewal_window_after_expiry_in_days.days
+  end
 
   def label_and_value(label, value)
     label_text = I18n.t("renewal_letter.renewal_letter_content.section_4.#{label}")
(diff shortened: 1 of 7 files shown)


app/presenters/notify_renewal_letter_presenter.rb
```diff
@@ -6,7 +6,11 @@ class NotifyRenewalLetterPresenter < BasePresenter
 
   def expiry_date
     # Currently you can only add exemptions when you register, so we can assume they expire at the same time
-    registration_exemptions.first.expires_on.to_formatted_s(:day_month_year)
+    first_exemption.expires_on.to_formatted_s(:day_month_year)
+  end
+
+  def renewal_window_start_date
+    (first_exemption.expires_on - renewal_window_after_expiry_in_days).to_formatted_s(:day_month_year)
```
timstone: don't you mean `renewal_window_before_expiry_in_days` ?

config/initializers/waste_exemptions_engine.rb
```diff
@@ -37,7 +37,7 @@
   configuration.use_last_email_cache = ENV["USE_LAST_EMAIL_CACHE"] || "false"
 
   # Renewing config
-  configuration.renewal_window_before_expiry_in_days = ENV["RENEWAL_WINDOW_BEFORE_EXPIRY_IN_DAYS"] || 28
+  configuration.renewal_window_before_expiry_in_days = ENV["RENEWAL_WINDOW_BEFORE_EXPIRY_IN_DAYS"] || 35
```
timstone: We don't want to change the renewal window, it should be 28 days

config/application.rb
```diff
@@ -67,7 +67,7 @@ class Application < Rails::Application
     config.epr_export_filename = ENV["EPR_DAILY_REPORT_FILE_NAME"] || "waste_exemptions_epr_daily_full"
     config.export_batch_size = ENV["EXPORT_SERVICE_BATCH_SIZE"] || 1000
 
-    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 35
+    config.ad_letters_exports_expires_in = ENV["AD_LETTERS_EXPORT_EXPIRES_IN"] || 33
```
timstone: I think the suggestion was to make this a dynamic number as there could be 28 days in the month, so one month + 3 days.

timstone (commented):
I won't appove PR's but that makes more sense to me now 👍

spec/services/notify_renewal_letter_service_spec.rb
```diff
@@ -3,24 +3,23 @@
...
+  describe ".run" do
     let(:registration) { create(:registration) }
-    let(:service) do
-      NotifyRenewalLetterService.run(registration: registration)
-    end
+    before { registration.registration_exemptions.update_all(expires_on: Date.new(2030, 7, 1)) }
 
     it "sends a letter" do
-      VCR.use_cassette("notify_renewal_letter") do
+      VCR.use_cassette("notify_renewal_letter", record: :new_episodes) do
```
PaulDoyle-EA: Minor point and not a blocker, but we shouldn't have to specify in an individual test that VCR can overwrite an existing cassette. The configuration might need some attention.

PaulDoyle-EA (dismissed):

spec/services/notify_renewal_letter_service_spec.rb
d-a-v-e: ah yeah, this is no longer needed.

PaulDoyle-EA (approved):

PR #856, opened 2022-06-01
Success banner and failure banner added for CH refresh button

Beckyrose200:
https://eaflood.atlassian.net/jira/software/projects/RUBY/boards/374?selectedIssue=RUBY-1877

Here we have changes for the CH back office refresh button. When the name refresh has been successful the user is informed with a success banner, likewise if it fails the user is told in the same manner.

When the refresh is unsuccessful the previous registered name in the db is kept and not overwritten with nil.


config/locales/refresh_companies_house_name.en.yml
```diff
@@ -0,0 +1,6 @@
+en:
+  refresh_companies_house_name:
+    messages:
+      success: You have successfully refreshed the Companies House information.
+      failure: There was an error refreshing the Companies House information. Try again later
+      failure_details: ""
```
PaulDoyle-EA: Add newline at end of file

PaulDoyle-EA (dismissed):

PaulDoyle-EA (changes requested):
As per my comment on the engine change, I think RefreshCompaniesHouseService should be in this repo, along with some service-level unit tests.

spec/services/refresh_companies_house_name_service_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      let(:old_registered_name) { nil }
+
+      it "stores the companies house name" do
+        expect { subject }.to change { registration_data(registration).operator_name }
+          .from(nil)
+          .to(new_registration_name)
+      end
+    end
+
+    context "with an existing registered company name" do
```
PaulDoyle-EA, line 51: Please add one more test here to ensure that an existing companies house name is not clobbered if the API call fails.

PaulDoyle-EA (approved):

PR #877, opened 2022-07-07
Map assistance_mode values for boxi registrations export

PaulDoyle-EA:
This change modifies the existing export of registration `assistance_mode` to meaningful text values for BOXI reports purposes.
https://eaflood.atlassian.net/browse/RUBY-1989

### DIFF
--- a/spec/models/reports/boxi/registrations_serializer_spec.rb
+++ b/spec/models/reports/boxi/registrations_serializer_spec.rb
@@ -22,27 +22,19 @@ module Reports
       end
 
       describe "#parse_assistance_mode" do
-        # let(:registration) { create(:registration) }
-
         context "for an unassisted registration" do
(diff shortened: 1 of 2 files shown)


spec/models/reports/boxi/registrations_serializer_spec.rb
```diff
@@ -20,6 +20,35 @@ module Boxi
           expect(result_lines.count).to eq(2)
         end
       end
+
+      describe "#parse_assistance_mode" do
+        # let(:registration) { create(:registration) }
```
d-a-v-e: are these still needed?

PaulDoyle-EA: Oops good catch

d-a-v-e (approved):

PR #880, opened 2022-07-11
Renew link, companies house data retrieval

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-1993

Here we have the routing changes to the back office renewal link button. This change allows a company or llp renewing to be redirected to the check_registered_name_and_address_form to validate their company name and address before being shown the check_your_answers+page in the renewal. For non companies the journey is still the same

### DIFF
--- a/spec/requests/renews_spec.rb
+++ b/spec/requests/renews_spec.rb
@@ -1,9 +1,19 @@
 # frozen_string_literal: true
 
 require "rails_helper"
+require "defra_ruby_companies_house"
 
 RSpec.describe "Renews", type: :request do
   let(:registration) { create(:registration) }
+  let(:transient_registration_token) { WasteExemptionsEngine::RenewingRegistration.last.token }
+  let(:company_name) { Faker::Company.name }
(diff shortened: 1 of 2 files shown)


spec/requests/renews_spec.rb
```diff
@@ -1,9 +1,19 @@
 # frozen_string_literal: true
 
 require "rails_helper"
+require "defra_ruby_companies_house"
 
 RSpec.describe "Renews", type: :request do
   let(:registration) { create(:registration) }
+  let(:transient_registration_token) { WasteExemptionsEngine::RenewingRegistration.last.token }
+  let(:company_name) { Faker::Company.name }
```
PaulDoyle-EA: I thunk lines 9-16 belong in the "when the business type is a company or llp" context block. Are those values used anywhere else?

Beckyrose200: They are used during the check for the user as well. Have changed them though to be inside that block

PaulDoyle-EA (approved):

PR #884, opened 2022-07-18
Send letters to contacts with no email address

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-1997

Here we have the changes to allow letters to be sent to registrations without an email address.
In WEX there are two emails address captured, applicant_email and contact_email.
These changes mean letters are sent to any emails that are nil value instead of the NCCC email address.


db/schema.rb
```diff
@@ -204,6 +204,8 @@
     t.string "temp_reuse_address_for_site_location"
     t.boolean "temp_use_registered_company_details"
     t.datetime "companies_house_updated_at"
+    t.boolean "temp_reuse_applicant_name"
```
PaulDoyle-EA: Please check why this change crept in. Possibly related to ongoing issues around WEX migrations.

Beckyrose200: The `temp_reuse_applicant_name` change was merged into the engine on July 11th. The back office schema hasn't been updated with this change since the merge. I ran a db:migrate due to issues with my db so I think that's why this PR has the updated schema.

Beckyrose200: I think the front office needs the schema updating as well

app/controllers/resend_confirmation_email_controller.rb
```diff
@@ -41,7 +41,7 @@ def send_emails
 
   def emails_to_contact
     emails = [registration.applicant_email, registration.contact_email]
-    emails.delete(WasteExemptionsEngine.configuration.assisted_digital_email)
+    emails.compact!
     emails.uniq
   end
```
d-a-v-e, line 46: looks like you can lose the assignment here, so you could just write (note the removal of the `!` in `compact!`):

```ruby
def emails_to_contact
  [registration.applicant_email, registration.contact_email]
    .delete(WasteExemptionsEngine.configuration.assisted_digital_email)
    .compact
    .uniq
end
```

app/services/remove_nccc_email_service.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# rubocop:disable Layout/LineLength
+class RemoveNcccEmailService < ::WasteExemptionsEngine::BaseService
+
+  # https://rubular.com/r/LPHA79nxD5k2ff
```
d-a-v-e: this is really handy. (Make sure it is cleared on rubular.com once this is merged!)

app/services/remove_nccc_email_service.rb
```diff
@@ -0,0 +1,35 @@
...
+    Rails.logger.info "Clearing wex applicant email addresses: #{matching_registrations.pluck(:applicant_email)}"
+    matching_registrations.update_all(applicant_email: nil)
+
+    matching_registrations = find_email_variations(:contact_email)
+
+    return unless matching_registrations.any?
+
+    Rails.logger.info "Clearing wex contact email addresses: #{matching_registrations.pluck(:contact_email)}"
+    matching_registrations.update_all(contact_email: nil)
+  end
```
d-a-v-e, line 33: So if there are no matches found for `:applicant_email` then the rest of the searches have no reason to run?

d-a-v-e (changes requested):

app/services/remove_nccc_email_service.rb
Beckyrose200: This was another annoying rubocop error. I had them in if statements and rubocop disagreed.

app/controllers/resend_confirmation_email_controller.rb
Beckyrose200: That looks good thank you

app/models/waste_exemptions_engine/registration.rb
```diff
@@ -18,7 +16,7 @@ class Registration < ::WasteExemptionsEngine::ApplicationRecord
 
     scope :renewals, -> { where.not(referring_registration_id: nil) }
 
-    scope :contact_email_is_not_nccc, -> { where.not(contact_email: NCCC_EMAIL) }
+    scope :contact_email_is_not_blank, -> { where.not(contact_email: nil) }
```
PaulDoyle-EA: "blank" and "nil" are not the same - maybe rename this to `contact_email_present` and also check for empty string values?

app/services/bulk_notify_renewal_letters_service.rb
```diff
@@ -24,7 +24,7 @@ def ad_expiring_registrations
     @_ad_expiring_registrations ||= lambda do
       WasteExemptionsEngine::Registration
         .order(:reference)
-        .where(contact_email: WasteExemptionsEngine.configuration.assisted_digital_email)
+        .where(contact_email: nil)
```
PaulDoyle-EA, line 27: Might need to also check for empty strings

app/services/remove_nccc_email_service.rb
PaulDoyle-EA: ... in which case the comment should be deleted ....

PaulDoyle-EA (approved):

d-a-v-e (approved):

PR #904, opened 2022-08-12
Add renewals column to boxi exemptions export

PaulDoyle-EA:
This change adds a "renewals" column to the Boxi `registration_exemptions` export.
https://eaflood.atlassian.net/browse/RUBY-1978


Beckyrose200 (approved):
👍

PR #911, opened 2022-08-18
Adding telephone to the search criteria

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-2057 
Here we have the changes for adding in applicant and contact telephone to the search results on the dashboard. 
This is only for an exact match and not a partial match

### DIFF
--- a/spec/factories/new_registration.rb
+++ b/spec/factories/new_registration.rb
@@ -34,6 +34,10 @@ FactoryBot.define do
       "0123456789#{n}"
     end
 
+    sequence :applicant_phone do |n|
+      "0123456789#{n}"
+    end
+
     transient_addresses do
       [build(:transient_address, :operator),
        build(:transient_address, :contact),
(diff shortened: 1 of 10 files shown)


spec/factories/new_registration.rb
```diff
@@ -30,6 +30,14 @@
       "Operator #{n}"
     end
 
+    sequence :contact_phone do |n|
```
PaulDoyle-EA, line 33: Are these still needed / used by the specs?

spec/models/new_registration_spec.rb
```diff
@@ -30,6 +30,26 @@
       end
     end
 
+    context "when the search is a phone number" do
+      let(:normal_number) { "01234567890" }
```
PaulDoyle-EA: These `let` statements could possibly be moved into the shared_example

PaulDoyle-EA (dismissed):

PaulDoyle-EA (approved):

PR #918, opened 2022-09-07
Feature/site grid reference

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-1931

Here we have added site grid reference into the site location section on the displayable registration page. 
We have added a header to the site address so its clear which is the address and which is the grid reference. 
The grid reference has logic in it so it will only show for the site location and not for the contact or registered location sections. 
Even if the grid reference is nil the header will still be shown as per the A\C. This alerts the BO user that the grid reference hasn't been successfully created.

### DIFF
--- a/app/views/shared/_resource_address.html.erb
+++ b/app/views/shared/_resource_address.html.erb
@@ -7,7 +7,10 @@
     <% end %>
   </dt>
   <dd class="govuk-summary-list__value">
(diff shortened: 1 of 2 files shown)


Cruikshanks (commented):
perfection

Cruikshanks (approved):

PR #919, opened 2022-09-08
Feature/postcode search for all address

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-2058

Here we have the changes for adding in contact and operator postcode to the BO search functions

### DIFF
--- a/spec/models/address_spec.rb
+++ b/spec/models/address_spec.rb
@@ -42,7 +42,7 @@ RSpec.describe WasteExemptionsEngine::Address, type: :model do
   describe "#contact" do
     let(:scope) { WasteExemptionsEngine::Address.contact }
 
-    it "returns site addresses" do
+    it "returns contact addresses" do
       expect(scope).to include(matching_address_contact)
     end
 
@@ -54,7 +54,7 @@ RSpec.describe WasteExemptionsEngine::Address, type: :model do
   describe "#operator" do
     let(:scope) { WasteExemptionsEngine::Address.operator }
 
-    it "returns site addresses" do
+    it "returns operator addresses" do
       expect(scope).to include(matching_address_operator)
     end
 
--- a/spec/models/address_spec.rb
+++ b/spec/models/address_spec.rb
@@ -3,8 +3,10 @@
 require "rails_helper"
 
 RSpec.describe WasteExemptionsEngine::Address, type: :model do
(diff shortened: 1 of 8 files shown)


spec/models/address_spec.rb
```diff
@@ -29,7 +31,31 @@
...
+
+    it "does not return others" do
+      expect(scope).not_to include(non_matching_address)
+    end
+  end
+
+  describe "#contact" do
+    let(:scope) { WasteExemptionsEngine::Address.contact }
+
+    it "returns site addresses" do
```
irisfaraway:
```suggestion
    it "returns contact addresses" do
```

spec/models/address_spec.rb
```diff
@@ -29,7 +31,31 @@
...
+
+    it "does not return others" do
+      expect(scope).not_to include(non_matching_address)
+    end
+  end
+
+  describe "#operator" do
+    let(:scope) { WasteExemptionsEngine::Address.operator }
+
+    it "returns site addresses" do
```
irisfaraway:
```suggestion
    it "returns operator addresses" do
```

spec/models/new_registration_spec.rb
```diff
@@ -166,11 +166,15 @@
         end
       end
 
-      context "when the address is not a site address" do
-        let(:address) { matching_registration.transient_addresses.first }
+      context "when the address is a contact address" do
```
irisfaraway, line 168: I think these three "when the address is a X address" tests may need some differences in their contexts – as it is, it just looks like the same test three times

spec/models/waste_exemptions_engine/registration_spec.rb
```diff
@@ -279,11 +279,15 @@
...
-        let(:address) { matching_registration.addresses.first }
+      context "when the address is a contact address" do
+        it "is included in the scope" do
+          expect(scope).to include(matching_registration)
+        end
+      end
 
-        it "is not included in the scope" do
-          expect(scope).to_not include(matching_registration)
+      context "when the address is an operator address" do
```
irisfaraway, line 289: See comment about identical tests above!

irisfaraway (changes requested):
Looking good – I just have some questions about the tests

irisfaraway (approved):
:shipit:

PR #972, opened 2022-12-06
Persist version on cease/revoke of registration exemptions

endofunky:


app/services/deregistration_service.rb
```diff
@@ -2,24 +2,27 @@
...
+    if resource.is_a?(WasteExemptionsEngine::Registration)
+      resource.registration_exemptions.each do |re|
+        DeregistrationService.new(current_user, re).deregister!(state_transition, deregistration_message)
       end
     else
-      # Apply the new state via the AASM helper method.
-      @resource.public_send("#{state_transition}!")
-      @resource.update(deregistration_message: deregistration_message)
+      # Assign current_user's email address as version author
+      PaperTrail.request(whodunnit: current_user.email) do
```
PaulDoyle-EA, line 21: Is the whodunnit already handled here? https://github.com/DEFRA/waste-exemptions-engine/blob/cf8f155cef878e14e6219985a47b7a580724d75a/app/controllers/waste_exemptions_engine/application_controller.rb#L21

PaulDoyle-EA (approved):

app/services/deregistration_service.rb
endofunky: Yes, but I wanted it to be explicit as to what we set as the `whodunnit` field in that service for future readers, and in case we ever have to call it outside of a controller context. Also, this allows testing of the service in isolation and avoids having to test this through request/controller specs.

PR #1030, opened 2023-03-09
Add resend deregistration email link

endofunky:


app/services/resend_deregistration_email_service.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    WasteExemptionsEngine::Registration.transaction do
+      registration.update!(deregistration_email_sent_at: nil)
+      SendDeregistrationEmailJob.perform_later(registration.id)
+    end
+
+    true
+  rescue StandardError => e
+    Airbrake.notify e
+    Rails.logger.error "Resend deregistration email for registration #{registration.id}:\n#{e}"
```
PaulDoyle-EA, line 17: `registration.reg_identifier` might be more useful for debugging; also send to Airbrake

spec/services/resend_deregistration_email_service_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    before do
+      allow(Airbrake).to receive(:notify)
+    end
+
+    context "when successful" do
+      it "resets the deregistration_email_sent_at timestamp" do
+        run_service
+
+        expect(registration.reload.deregistration_email_sent_at).to be_nil
```
PaulDoyle-EA, line 33: Why is this set to nil if the email is sent?

endofunky: The job that sends the de-registration email checks if the timestamp is `nil` in order to not send it multiple times (for example, resend button was clicked more than once and two jobs are enqueued to send the email). So in order to make the job send the email we have to send the timestamp back to `nil`, which is what this example asserts. The job hasn't executed in the context of the spec (we just check if it was enqueued) and hence we just check that the timestamp has correctly been set to `nil`.

app/services/resend_deregistration_email_service.rb
endofunky: Updated with 2d9da93

endofunky: Sorry, amended commit once more to 122c541 .. this is called `reference` instead of `reg_identified`

PaulDoyle-EA (approved):

PR #1089, opened 2023-06-26
Feature/ruby 2455 wex notify text messaging renewal reminder texts

brujeo:
https://eaflood.atlassian.net/browse/RUBY-2455

### DIFF
--- a/app/models/waste_exemptions_engine/registration.rb
+++ b/app/models/waste_exemptions_engine/registration.rb
@@ -27,6 +27,8 @@ module WasteExemptionsEngine
 
     scope :contact_email_present, -> { where.not(contact_email: nil) }
 
+    scope :applicant_phone_present, -> { where.not(applicant_phone: nil) }
+
     scope :site_address_is_not_nccc, lambda {
       joins(:addresses).merge(Address.site.not_nccc)
     }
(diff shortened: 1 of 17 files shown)


app/models/waste_exemptions_engine/registration.rb
```diff
@@ -58,5 +60,11 @@ def renewable?
     def in_renewable_state?
       active? || expired?
     end
+
+    def valid_mobile_phone_number?
+      record = clone
```
PaulDoyle-EA, line 65: What is this cloning?

app/services/final_renewal_reminder_service.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class FinalRenewalReminderService < RenewalReminderTextServiceBase
```
PaulDoyle-DEFRA, line 3: I found during current work across multiple communications services that it is helpful to have the communications type in the service name - so maybe "FinalRenewalReminderTextService"

PaulDoyle-EA (approved):

jjromeo (approved):

app/models/waste_exemptions_engine/registration.rb
brujeo: I'm reusing active record validator here which populates the instance error messages list with an error. My worry is that if I do that on an actual instance, this might cause issues while updating user details (because the error list is not empty).  This method does not suppose to perform validation, but just to check if phone number is valid mobile phone.

PR #1166, opened 2023-10-10
Fix/ruby 2735 clear transient reg

PaulDoyle-EA:
This change removes any existing work-in-progress transient_registrations when a renewal is started in the back office. This is to avoid occasional issues in production where a transient registration is badly formed and this prevents the user from completing renewal.
https://eaflood.atlassian.net/browse/RUBY-2735


app/controllers/renews_controller.rb
```diff
@@ -6,6 +6,10 @@ class RenewsController < ApplicationController
   def new
     authorize
 
+    # Back office only: Badly formed transient registrations sometimes cause renewals to fail.
+    # So we clear any existing transient registrations before starting a back-office renewal.
+    WasteExemptionsEngine::TransientRegistration.where(reference: registration.reference).destroy_all
```
brujeo, line 11: Is this approach safe? I've imagined a situation where I have two windows open. I've almost completed the renewal in one window, and then accidentally clicked renew in another. Will the renewal that is in progress get removed as a result?

timstone: Yes I think it's sound.

PR #1192, opened 2023-11-08
Ruby 2742 wex bo add reset transient registration action to registrations

jjromeo:


lib/tasks/cleanup.rake
```diff
@@ -2,8 +2,9 @@
 
 namespace :cleanup do
   desc "Remove old transient_registrations from the database"
-  task transient_registrations: :environment do
-    TransientRegistrationCleanupService.run
+  task :transient_registrations, [:limit] => [:environment] do |_task, args|
+    limit = args[:limit]
```
PaulDoyle-EA: This will need to be parsed from string to integer

spec/lib/tasks/cleanup_spec.rb
```diff
@@ -6,9 +6,26 @@
...
 
   describe "cleanup:transient_registrations" do
+    before do
+      @old_registration = create(:new_registration, created_at: 31.days.ago)
+      create(:new_registration, created_at: 31.days.ago)
+      create(:new_registration)
+    end
+
+    it "deletes only up to the limit number of old transient registrations" do
+      expect { subject.invoke(1) }.to change {
```
PaulDoyle-EA: I think this should be tested using a string value

PR #1243, opened 2024-01-26
Add one-off rake task to fix communication_log data anomalies

PaulDoyle-EA:
https://eaflood.atlassian.net/browse/RUBY-2852

### DIFF
--- a/spec/lib/tasks/one_off/fix_communication_log_anomalies_spec.rb
+++ b/spec/lib/tasks/one_off/fix_communication_log_anomalies_spec.rb
@@ -25,22 +25,38 @@ RSpec.describe "one_off:fix_communication_log_anomalies", type: :rake do
   it { expect { run_rake_task }.not_to raise_error }
 
   it {
-    expect { run_rake_task }
-      .to change {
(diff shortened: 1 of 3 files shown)


spec/lib/tasks/one_off/fix_communication_log_anomalies_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    create(:communication_log, template_id: registration_edit_link_email_id, template_label: "Registration edit link email")
+    create(:communication_log, template_id: registration_edit_link_email_id, template_label: "Update your contact details and deregister your waste exemptions")
+  end
+
+  # By default Rails prevents multiple invocations of the same Rake task in succession
+  after { rake_task.reenable }
+
+  it { expect { run_rake_task }.not_to raise_error }
+
+  it {
```
jjromeo, line 27: This spec only tests for the removal of communication log anomalies, rather than for if they are fixed

PaulDoyle-EA: Good point

jjromeo (approved):

PR #1299, opened 2024-04-17
[RUBY-3041] rake task to fix user journeys incorrectly marked as incomplete

brujeo:
https://eaflood.atlassian.net/browse/RUBY-3041

### DIFF
--- a/lib/tasks/one_off/fix_incomplete_user_journeys.rake
+++ b/lib/tasks/one_off/fix_incomplete_user_journeys.rake
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+namespace :one_off do
+  # https://eaflood.atlassian.net/browse/RUBY-3041
+  desc "Fix user journeys incorrectly marked as incomplete"
+  task fix_incomplete_user_journeys: :environment do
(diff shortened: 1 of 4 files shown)


lib/tasks/one_off/fix_incomplete_user_journeys.rake
```diff
@@ -0,0 +1,24 @@
...
+  task fix_incomplete_user_journeys: :environment do
+    incomplete_journeys = WasteExemptionsEngine::Analytics::UserJourney.where(completed_route: nil)
+    incomplete_journeys.each do |journey|
+      # check if journey passed the completion route
+      journey.page_views.each do |page_view|
+        page = page_view.page
+        next unless WasteExemptionsEngine::Analytics::UserJourney::COMPLETION_PAGES.include?(page)
+
+        # complete the journey
+        journey.update(
```
PaulDoyle-EA, line 15: Shouldn't this call `journey.complete_journey` instead?

brujeo: complete_journey requires transient_registration as a parameter and cannot be called retrospectively. So, I had to come up with different way to mark journeys as complete

PaulDoyle-EA (approved):

PR #1313, opened 2024-04-29
[RUBY-2749] Add deregistration details page, and enhance styles

jjromeo:
- Introduce DeregistrationsController with a show action to display deregistration details.
- Extend RegistrationExemption model with a deregistered_by method to track the user who performed the deregistration.
- Create a new view for showing deregistration details in deregistrations/show.html.erb.
- Update the resource exemptions list partial to include links to the new deregistration details page for ceased or revoked exemptions.
- Add localization entries for deregistration details in config/locales/deregistrations.en.yml.
- Include a new route for deregistrations show action.
- Add tests for the deregistered_by method in registration_exemption_spec.rb to ensure proper functionality.
- Add styling for .govuk-table__caption--m in application.scss to align the table caption to the left.

This commit introduces the ability to view deregistration details for exemptions and includes necessary model, view, controller, and styling changes to support this feature. It also ensures that the Gemfile points to the correct feature branch for the waste_exemptions_engine gem.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2024_03_20_151621) do
+ActiveRecord::Schema[7.1].define(version: 2024_03_20_151621) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
   enable_extension "tsm_system_rows"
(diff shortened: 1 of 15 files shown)


db/schema.rb
```diff
@@ -151,9 +151,10 @@
     t.datetime "deregistration_email_sent_at", precision: nil
     t.string "edit_token"
     t.datetime "edit_token_created_at"
-    t.boolean "reminder_opt_in", default: true
```
PaulDoyle-EA: Why has this changed?

jjromeo: Hmm, I'm not sure and it certainly should not have

PaulDoyle-EA (approved):

PR #1314, opened 2024-04-30
RUBY-3050: WEX Charging: Band and charge management

brujeo:
https://eaflood.atlassian.net/browse/RUBY-3050
* Band management
* Registration charge management
* Farmer bucket management


app/controllers/bands_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def edit
+    find_band(params[:id])
+  end
+
+  def create
+    @band = WasteExemptionsEngine::Band.new(band_params)
+    # copy registration_charge from the first band
+    @band.registration_charge = if WasteExemptionsEngine::Band.count.positive?
```
jjromeo: Does this code need to be in the controller?

app/controllers/bands_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    find_band(params[:id])
+
+    if @band.update(band_params)
+      redirect_to bands_url
+    else
+      render :edit
+    end
+  end
+
+  def edit_registration_charge
```
jjromeo: This is only food for thought but would another controller make sense for the registration_charge actions

app/views/bands/index.html.erb
```diff
@@ -0,0 +1,98 @@
...
+      </table>
+    </div>
+  </div>
+<% end %>
+
+
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <table class="govuk-table" aria-label="<%= t(".heading") %>">
```
jjromeo, line 51: It will now say "Band and charges management" twice on the page?

app/views/bands/index.html.erb
```diff
@@ -0,0 +1,98 @@
...
+        </thead>
+        <tbody class="govuk-table__body">
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell">
+              <%= t('.other_charges.list.registration_charge') %>
+            </td>
+            <td class="govuk-table__cell">
+            </td>
+            <td class="govuk-table__cell">
+              <%= display_pence_as_pounds_and_cents(@bands&.first&.registration_charge, hide_pence_if_zero: false) %>
```
jjromeo: @bands should always be an active record relation so shouldn't need the first safe navigation operator?

config/routes.rb
```diff
@@ -63,6 +63,14 @@
   # Privacy policy
   get "/ad-privacy-policy", to: "ad_privacy_policy#show", as: :ad_privacy_policy
 
+  # Charges management: bands, exemptions, buckets and charges
+  resources :bands, only: %i[index new create edit update] do
```
jjromeo: As these are the majority, maybe except: :destroy would be better

app/controllers/bands_controller.rb
brujeo: this has been fully refactored now

app/controllers/bands_controller.rb
brujeo: This has been fully refactored now

app/views/bands/index.html.erb
brujeo: nope, aria-label is an attribute for screen readers

app/views/bands/index.html.erb
brujeo: This has been fully refactored now

config/routes.rb
brujeo: I prefer them to be listed explicitly. In any case, this only: condition will be gone, once you'll merge in your charges for band deletion feature

app/controllers/bands_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+  def band_params
+    params.require(:band)
+          .permit(
+            :name,
+            :sequence,
+            :registration_charge,
+            initial_compliance_charge_attributes: %i[id charge_amount_in_pounds],
+            additional_compliance_charge_attributes: %i[id charge_amount_in_pounds]
+          ).tap do |params|
+      if params[:initial_compliance_charge_attributes]
```
jjromeo, line 62: I guess this is for the creation only, would there be another place to do this rather than inside the params?

brujeo: These are required attributes that have to be added but don't exist in the current interface. In theory, I could have preset them in the form of hidden fields but this way of setting is just simpler. I don't have a strong opinion on this, so if you'll suggest any good way of refactoring it - I can do it

jjromeo (dismissed):

jjromeo (approved):

PR #1328, opened 2024-05-09
RUBY 3065 wex charging bo band charges deletion

jjromeo:
[Jira link](https://eaflood.atlassian.net/browse/RUBY-3065)

This pr introduces the ability to delete bands from the system with a confirmation step. A new view `destroy_confirmation.html.erb` is added, which asks the user to confirm the deletion of a band. The `BandsController` is updated with a `destroy_confirmation` action to render the confirmation page and a `destroy` action to handle the actual deletion. The deletion process includes a check for any exemptions associated with the band; if exemptions exist, the user is redirected to a new `cannot_destroy.html.erb` view that informs them that bands with allocated exemptions cannot be deleted.

The `index.html.erb` view for bands is updated to replace the placeholder delete link with a link to the new `destroy_confirmation` action. The routes file is also updated to include the new `destroy_confirmation` and `cannot_destroy` member routes within the bands resource block.

Additionally, request specs are added to verify the new `destroy_confirmation` page's functionality, ensuring it renders correctly when a permitted user is signed in.

Files added:
- `app/views/bands/cannot_destroy.html.erb`
- `app/views/bands/destroy_confirmation.html.erb`

Files modified:
- `app/controllers/bands_controller.rb`: Added `destroy_confirmation` and `destroy` actions.
- `app/views/bands/index.html.erb`: Updated the delete link to point to the confirmation action.
- `config/routes.rb`: Added routes for `destroy_confirmation` and `cannot_destroy`.
- `spec/requests/bands_spec.rb`: Added specs for the `destroy_confirmation` action.
- `config/locales/bands.en.yml`: Added text for new html views

This change ensures that bands can be removed from the system in a user-friendly and safe manner, preventing accidental deletions and maintaining data integrity when exemptions are involved.

### DIFF
--- a/app/controllers/bands_controller.rb
+++ b/app/controllers/bands_controller.rb
@@ -74,7 +74,6 @@ class BandsController < ApplicationController
           .permit(
             :name,
             :sequence,
-            :registration_charge,
             initial_compliance_charge_attributes: %i[id charge_amount_in_pounds],
             additional_compliance_charge_attributes: %i[id charge_amount_in_pounds]
           ).tap do |params|
--- a/app/controllers/bands_controller.rb
+++ b/app/controllers/bands_controller.rb
@@ -56,7 +56,7 @@ class BandsController < ApplicationController
       flash[:message] = I18n.t("bands.destroy.success")
       redirect_to bands_url
     else
-      flash[:error] = I18n.t("bands.destroy.error")
+      flash[:error] = I18n.t("bands.destroy.failure")
       render :index
     end
   end
--- a/app/controllers/bands_controller.rb
+++ b/app/controllers/bands_controller.rb
@@ -76,7 +76,6 @@ class BandsController < ApplicationController
           .permit(
             :name,
             :sequence,
-            :registration_charge,
             initial_compliance_charge_attributes: %i[id charge_amount_in_pounds],
             additional_compliance_charge_attributes: %i[id charge_amount_in_pounds]
           ).tap do |params|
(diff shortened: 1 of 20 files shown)


spec/requests/bands_spec.rb
```diff
@@ -190,4 +190,65 @@
...
+      before do
+        sign_in(user)
+      end
+
+      context "when the band has exemptions" do
+        before do
+          create(:exemption, band: band)
+        end
+
+        it "redirects to the cannot destroy page" do
```
PaulDoyle-EA, line 222: Maybe add a spec to confirm that it does not delete the band.

PaulDoyle-EA (dismissed):

app/controllers/bands_controller.rb
```diff
@@ -39,6 +39,26 @@ def update
...
+  end
+
+  def cannot_destroy
+    # This is a placeholder method to render the cannot_destroy template
+  end
+
+  def destroy
+    find_band(params[:id])
+
+    if @band.exemptions.any?
```
brujeo: I'd expect a dedicated method for checking whether the band can be deleted, something like @band.can_be_destroyed? also I'd also add a callback to execute this validation before destroy is triggered. This is to avoid accidental deletions (e.g. by cascading foreign key or some service that doesn't use this controller).

But as we discussed that previously, it makes sense to keep it simple for now till all requirements are clear

brujeo (dismissed):

brujeo (approved):

PR #1514, opened 2024-11-06
RUBY-2705: WEX charging - Add a payment page

brujeo:
Add a payment page
https://eaflood.atlassian.net/browse/RUBY-2705


app/forms/add_payment_form.rb
```diff
@@ -0,0 +1,66 @@
...
+      payment_status: WasteExemptionsEngine::Payment::PAYMENT_STATUS_SUCCESS,
+      date_time: Date.new(date_year.to_i, date_month.to_i, date_day.to_i),
+      payment_uuid: SecureRandom.uuid,
+      reference: payment_reference,
+      account: account,
+      comments: comments
+    )
+  end
+
+  def validate_payment_date_current_or_past
```
PaulDoyle-EA, line 58: `DefraRuby::Validators::PastDateValidator` does this and includes a guard against two-digit year entry (year < 1900)

db/schema.rb
```diff
@@ -196,9 +196,11 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.string "payment_uuid"
+    t.bigint "account_id"
     t.string "reference"
+    t.string "comments", limit: 500
+    t.index ["account_id"], name: "index_payments_on_account_id"
     t.index ["order_id"], name: "index_payments_on_order_id"
-    t.index ["reference"], name: "index_payments_on_reference"
```
PaulDoyle-EA, line 201: Why remove this index?

brujeo: It's not removed, it's still there. Please check the full file
https://github.com/DEFRA/waste-exemptions-back-office/blob/93e020fc7a896a2589c873735df01b824c0a2e2e/db/schema.rb

I believe there was a duplicate line caused by merge conflict that I had to remove

app/forms/add_payment_form.rb
brujeo: suggested validator works well for a single date field. However, in my case, 3 separate integer fields are being submitted: day, month and year

PaulDoyle-EA (approved):

PR #1526, opened 2024-11-19
[RUBY-2704] Fix bug where refund can be higher than balance

jjromeo:
- Enhanced validation to check if refund amount is greater than or equal to the balance.
- Removed redundant balance assignment in test setup.
- Added a test case to ensure refund amount does not exceed balance.
- Refactor error messages to use the convention for displaying from en.yml
- set date in payments table to date_time to fix regression in table
- Modified the refund form to display dynamic payment types based on `@payment.payment_type`.
- Refactored form fields to use `govuk_text_field` and `govuk_text_area` to fix issue with errors not linking back to ifelds
- Updated locale file to include additional payment types and improve translation keys.


PaulDoyle-EA (dismissed):

app/forms/record_refund_form.rb
```diff
@@ -7,19 +7,21 @@ class RecordRefundForm
...
     self.amount = params[:amount]
     self.comments = params[:comments]
     self.payment_id = params[:payment_id]
 
+    return false unless check_payment_exists?
+
     @payment = WasteExemptionsEngine::Payment.find_by(id: payment_id)
-    @balance = payment&.account&.balance
+
+    @balance = payment.account.balance / 100
```
brujeo: It would make sense to use existing helper for price conversion
display_pence_as_pounds_and_cents
or maybe even add a convenience method 
`payment.account.balance_in_pounds`

app/forms/record_refund_form.rb
```diff
@@ -36,28 +38,20 @@ def submit(params, record_refund_service: RecordRefundService)
...
+    false
   end
 
   def amount_within_limits
     return if amount.blank? || !amount.to_f.positive?
 
-    if amount.to_f > payment.payment_amount.to_f / 100
-      errors.add(:amount, I18n.t(".record_refunds.form.errors.exceeds_payment_amount"))
-    end
+    errors.add(:amount, :exceeds_payment_amount) if amount.to_f > payment.payment_amount.to_f / 100
```
brujeo: same here,  consider using existing helper method for price conversion
`display_pence_as_pounds_and_cents `

app/forms/record_refund_form.rb
```diff
@@ -7,57 +7,55 @@ class RecordRefundForm
...
+    return false unless check_payment_exists?
+
     @payment = WasteExemptionsEngine::Payment.find_by(id: payment_id)
-    @balance = payment&.account&.balance
+
+    @balance = payment.account.balance
 
     return false unless valid?
 
     Rails.logger.info "running RecordRefundService with arguments: #{comments}, #{payment}, #{amount.to_f}"
```
brujeo: Is this log mesage still needed? This log message looks pretty useless without the context: account_id, payment_id, etc

brujeo (approved):

PR #1532, opened 2024-11-26
Chore/ruby 3465 remove super agent role

PaulDoyle-EA:
This change:
- Refactors roles and permissions definition to load from a CSV file instead of being hardcoded
- Refactors Ability unit tests to focus on groups of permissions instead of assuming a hierarchy of permissions
- Removes the redundant super_agent role
https://eaflood.atlassian.net/browse/RUBY-3465


config/initializers/roles_and_permissions.rb
```diff
@@ -0,0 +1,28 @@
...
+                  converters: ->(f) { f&.strip })
+
+  roles = csv.headers[1..]
+  role_permissions = roles.index_with { |_role| [] }
+
+  csv.each do |row|
+    roles.each do |role|
+      role_permissions[role] << row["permission"] if row[role].upcase == "Y"
+    end
+  end
```
brujeo: I think this code needs some refactoring to ensure:
- it doesn't crash the entire system if csv not present
- it doesn't crash the entire system if csv has typos or any other unexpected structure/data

brujeo (dismissed):

PR #1550, opened 2024-12-10
RUBY-3384: Add service_manager role

brujeo:
Add Service manager role
https://eaflood.atlassian.net/browse/RUBY-3384

### DIFF
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -71,7 +71,7 @@ RSpec.describe Ability do
     it_behaves_like "can manage users"
     it_behaves_like "can manage registrations"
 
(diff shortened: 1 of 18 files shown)


spec/models/ability_spec.rb
```diff
@@ -64,6 +64,17 @@
...
   end
 
+  context "when the user role is service_manager" do
+    let(:user) { build(:user, :service_manager) }
+
+    it_behaves_like "can use back office"
+    it_behaves_like "can manage users"
+    it_behaves_like "can manage registrations"
+
+    it { expect(ability).not_to be_able_to(:read, Reports::DefraQuarterlyStatsService) }
```
timstone: I'd like to double check this, it doesn't seem right.

PaulDoyle-EA (approved):

PR #1597, opened 2025-01-24
RUBY-3589: wex charging private beta send invite email

brujeo:
Ability to send private beta invite email
https://eaflood.atlassian.net/browse/RUBY-3589


app/controllers/new_registrations_controller.rb
```diff
@@ -12,7 +12,10 @@ def show
   private
 
   def find_resource(id)
-    @resource = WasteExemptionsEngine::NewRegistration.find(id)
+    @resource = WasteExemptionsEngine::NewRegistration.find_by(id: id)
```
PaulDoyle-EA, line 15: Why was this changed?

PaulDoyle-EA (approved):

app/controllers/new_registrations_controller.rb
brujeo: find(id) throws an error if not found, find_by(id: id) doesn't. So, this method was amended to look for NewChargedRigistration if NewRegistration is not found.

PR #1608, opened 2025-01-28
Include new_charged_registrations in "New registrations" search results

PaulDoyle-EA:
The dashboard search option for "New registrations" did not previously return any private beta transient registrations. This change searches on a subset of transient_registration classes so that both new_registrations and new_charged_registrations (but no other transient_registration types) will be returned in the search results.
https://eaflood.atlassian.net/browse/RUBY-3579

### DIFF
--- a/app/models/concerns/can_be_searched_like_registration.rb
+++ b/app/models/concerns/can_be_searched_like_registration.rb
@@ -3,8 +3,10 @@
 module CanBeSearchedLikeRegistration
   extend ActiveSupport::Concern
 
+  # rubocop:disable Metrics/BlockLength
   included do
-    scope :search_registration_and_relations, lambda { |term|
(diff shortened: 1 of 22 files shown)


app/models/concerns/can_be_searched_like_registration.rb
```diff
@@ -3,8 +3,10 @@
 module CanBeSearchedLikeRegistration
   extend ActiveSupport::Concern
 
+  # rubocop:disable Metrics/BlockLength
```
brujeo: I'm not a big fan of inline rubocop:disable comments. If it's a permanent thing, then better to move it to rubocop config and exclude this concern from Metrics/BlockLength checks

brujeo (approved):

PR #1641, opened 2025-02-13
Feature/ruby 3492 add support for farming exemptions

brujeo:
Daily Finance Data Export - added support for farming exemptions
https://eaflood.atlassian.net/browse/RUBY-3492


app/models/reports/finance_data_report/data_serializer.rb
```diff
@@ -45,48 +47,89 @@ def generate_order_rows(registration, order)
...
+        @total = 0
+
         # registration charge row
         charge_detail = order.charge_detail
         order_rows << registration_charge_row(registration, charge_detail)
+        order_rows.concat(generate_band_charge_rows(registration, charge_detail))
+        order_rows.concat(generate_farm_compliance_row(registration, charge_detail))
+        order_rows.concat(generate_charge_adjustment_rows(registration))
+        order_rows.concat(generate_payment_rows(registration))
+        order_rows
```
PaulDoyle-EA: Do we need to `.compact` this to remove any blank rows, e.g. if the registration has no farm exemptions?

app/models/reports/finance_data_report/data_serializer.rb
```diff
@@ -45,48 +47,89 @@ def generate_order_rows(registration, order)
...
+      end
+
+      def generate_band_charge_rows(registration, charge_detail)
+        rows = []
+        charge_detail.band_charge_details.each do |band_charge_detail|
+          if band_charge_detail.initial_compliance_charge_amount.positive?
```
PaulDoyle-EA: Minor: It might be tidier if these checks were moved inside the row methods (in this case `initial_compliance_charge_row` and let them return `[]` if the condition is false. Then the suggested use of `.compact` above would remove all empty rows.

spec/models/reports/finance_data_report/data_serializer_spec.rb
```diff
@@ -74,7 +74,29 @@ module FinanceDataReport
...
+            expect(csv[row_number]["charge_type"]).to eq("compliance_farm")
+            expect(csv[row_number]["charge_amount"]).to be_present
+            expect(csv[row_number]["charge_band"]).to be_nil
+            expect(csv[row_number]["exemption"]).to be_present
+            expect(csv[row_number]["payment_type"]).to be_nil
+            expect(csv[row_number]["refund_type"]).to be_nil
+            expect(csv[row_number]["reference"]).to be_nil
+            expect(csv[row_number]["comments"]).to be_nil
+            expect(csv[row_number]["payment_amount"]).to be_nil
+            expect(csv[row_number]["on_a_farm"]).to be_present
```
PaulDoyle-EA: I believe these two should `be true`, not just present

PaulDoyle-EA (dismissed):

app/models/reports/finance_data_report/data_serializer.rb
brujeo: .compact does it already to some degree. But yes, the use of .compact allowed to simplify the structure. Refactored

app/models/reports/finance_data_report/data_serializer.rb
brujeo: refactored

spec/models/reports/finance_data_report/data_serializer_spec.rb
brujeo: fixed

PaulDoyle-EA (dismissed):

PaulDoyle-EA (approved):

PR #1652, opened 2025-02-18
feature/RUBY-3636_start_new_charged_registration

PaulDoyle-EA:
Route from ad-privacy page to the charged registration journey if feature toggle is enabled
https://eaflood.atlassian.net/browse/RUBY-3636


app/helpers/ad_privacy_policy_helper.rb
```diff
@@ -13,6 +13,9 @@ def link_to_privacy_policy
   def destination_path
     if @registration.present?
       renew_path(reference: @registration.reference)
+    elsif WasteExemptionsEngine::FeatureToggle.active?(:back_office_private_beta_link)
```
timstone: could you call this feature toggle something else please? Could we move away from any PB reference?

spec/helpers/ad_privacy_policy_helper_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  describe "destination_path" do
+    context "when a registration is present" do
+      before { @registration = build(:registration) }
+
+      # rubocop:disable RSpec/InstanceVariable -- the helper checks for an instance variable
+      it "returns the renewal path for the registration" do
+        expect(helper.destination_path).to eq renew_path(reference: @registration.reference)
+      end
+      # rubocop:enable RSpec/InstanceVariable
```
brujeo: rubocop is rightly complaining about the use of instance variables.  It's easy just to rewrite the spec instead of having to disable rubocop rules
`
let(:registration) { build(:registration) }

it "returns the renewal path for the registration" do
    expect(helper.destination_path).to eq renew_path(reference: registration.reference)
end
`

PaulDoyle-EA: The instance variable is used in the helper, not in the spec.

brujeo (dismissed):

spec/helpers/ad_privacy_policy_helper_spec.rb
PaulDoyle-EA: As discussed - updated to remove the instance variable. Please re-review.

app/views/ad_privacy_policy/show.html.erb
```diff
@@ -57,7 +57,7 @@
     </details>
 
     <p class="govuk-body">
-      <%= link_to "Continue", destination_path, class: "govuk-button" %>
+      <%= link_to "Continue", destination_path(@reference), class: "govuk-button" %>
```
brujeo: I think you should pass registration not the reference

brujeo (approved):

PR #1664, opened 2025-02-21
[RUBY-3658] Refactor payment row presenter to handle negative payment amounts and calculate balance

brujeo:
Refactor payment row presenter to handle negative payment amounts and calculate balance
https://eaflood.atlassian.net/browse/RUBY-3658

### DIFF
--- a/app/presenters/reports/finance_data_report/payment_row_presenter.rb
+++ b/app/presenters/reports/finance_data_report/payment_row_presenter.rb
@@ -16,17 +16,17 @@ module Reports
       end
 
       def payment_amount
-        display_pence_as_pounds_and_pence(pence: record_amount, hide_pence_if_zero: true)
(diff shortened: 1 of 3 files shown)


app/presenters/reports/finance_data_report/payment_row_presenter.rb
```diff
@@ -16,11 +16,23 @@ def reference
...
 
       def balance
-        display_pence_as_pounds_and_pence(pence: @secondary_object.account.balance, hide_pence_if_zero: true)
+        @total -= amount_to_deduct
+        display_pence_as_pounds_and_pence(pence: @total, hide_pence_if_zero: true)
+      end
+
+      private
+
+      def amount_to_deduct
```
PaulDoyle-EA, line 29: This is called twice, should it be memoised?

PaulDoyle-EA (approved):

PR #1710, opened 2025-03-13
[RUBY-3675] Update renewal email to use `TemporaryFirstRenewalReminderService`

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-3675

Add a class `RenewalReminderEmailServiceSwitcher` to switch between the temporary first renewal reminder service and the renewal reminders, controlled by the `enable_renewals` feature flag. Additionally refactor the Temporary reminder emails to more closely resemble the structure of the actual renewal reminder emails.

### DIFF
--- a/app/services/renewal_service_switcher.rb
+++ b/app/services/renewal_service_switcher.rb
@@ -2,38 +2,38 @@
 
 class RenewalServiceSwitcher
   def self.first_reminder_service
-    if use_temporary_services?
-      TemporaryFirstRenewalReminderService
-    else
+    if renewals_enabled?
       FirstRenewalReminderService
+    else
+      TemporaryFirstRenewalReminderService
     end
   end
 
   def self.second_reminder_service
-    if use_temporary_services?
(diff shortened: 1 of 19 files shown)


app/services/renewal_service_switcher.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class RenewalServiceSwitcher
```
brujeo: I think that the service name does not reflect what the service is doing. It's not just switching but also sending out first renewal reminder emails

jjromeo: It is switching between the email services, would `RenewalReminderServiceSwitcher` make it more clear or any other suggestions?

brujeo: maybe FirstRenewalReminderSwitcherService ?

jjromeo: It contains several Renewal reminder services however, not just the First one

brujeo (approved):

PR #1732, opened 2025-03-28
Feature/ruby 3541 wex bau display registration changes audit trail

brujeo:
Registration Change History page
https://eaflood.atlassian.net/browse/RUBY-3541


.rubocop_todo.yml
brujeo, line 1: It seems that the RuboCop gem was previously upgraded, causing new offenses to be reported. This is unrelated to this PR.

app/services/registration_change_history_service.rb
```diff
@@ -21,32 +21,57 @@
...
+    contact_position
+    contact_phone
+    contact_email
+    on_a_farm
+    is_a_farmer
+  ].freeze
+
   def run(registration)
-    registration.versions.map do |version|
+    registration.versions.includes(:item).map do |version|
```
PaulDoyle-EA, line 45: Don't all versions have an `item` ?

brujeo: specs don't in some use cases. I was getting Bullet::Notification::UnoptimizedQueryError (eager loading detected) without this bt added

PaulDoyle-EA (approved):

PR #1838, opened 2025-07-24
[RUBY-3764] Update waste_exemptions_engine branch and improve EA Public Face Area data loading

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-3764
- Update gemfile to version of engine with ffi-geos
- Changed `EaPublicFaceAreaDataLoadService` to skip features marked as seaward (keep the land based features). The feature zip file contains duplicates of some of the regions which include an offshore and onshore version of some areas, before they were getting arbitrarily replaced based on the order of the file

### DIFF
--- a/app/services/ea_public_face_area_data_load_service.rb
+++ b/app/services/ea_public_face_area_data_load_service.rb
@@ -8,30 +8,31 @@ class EaPublicFaceAreaDataLoadService < WasteExemptionsEngine::BaseService
     features = RGeo::GeoJSON.decode(geo_json)
 
     ActiveRecord::Base.transaction do
-
-      features.each do |feature|
-        next if feature.properties["seaward"] == "Yes"
-
-        attributes = {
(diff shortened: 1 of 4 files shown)


app/services/ea_public_face_area_data_load_service.rb
```diff
@@ -8,29 +8,32 @@ def run
...
+      name: feature.properties["long_name"],
+      area_id: feature.properties["identifier"],
+      area: feature.geometry
+    }
+
+    area = WasteExemptionsEngine::EaPublicFaceArea.where(code: feature.properties["code"]).first
+
+    if area.present?
+      area.update(attributes)
+      puts "Updated EA Public Face Area \"#{area.code}\" (#{area.name})" unless Rails.env.test? # rubocop:disable Rails/Output
```
brujeo: Why do we need this output here?

brujeo (approved):

PR #1883, opened 2025-10-06
[RUBY-3997 & RUBY-3998 & RUBY-3991] wex bo multisite: registration sites page + exemptions details page + site exemptions deregistration

brujeo:
## RUBY-3997: WEX - Multiple site registrations - BO - sites page
https://eaflood.atlassian.net/browse/RUBY-3997

- Enhanced WasteExemptionsBackOffice::Address model with convenience methods such as site_status, reference, etc
- Implemented registration sites page that supports single-site and multi-site registrations

## RUBY-3998: WEX - Multiple site registrations - BO - exemptions details page
https://eaflood.atlassian.net/browse/RUBY-3998

- added new page to list site exemptions
 
## RUBY-3991: WEX - Multiple site registrations - BO - site exemptions deregistration
https://eaflood.atlassian.net/browse/RUBY-3991

- refactored existing deregistration page and related sections to support site deregistrations


app/views/shared/_resource_exemptions_list.html.erb
```diff
@@ -1,5 +1,7 @@
 <table class="govuk-table">
+  <% unless defined?(hide_caption) && hide_caption %>
```
PaulDoyle-EA, line 2: Maybe just `hide_caption.present?` ?

brujeo: I'm pretty sure this will raise an error like "undefined local variable or method 'hide_caption'..." when hide_caption is not defined

app/presenters/deregistrations_presenter.rb
```diff
@@ -38,8 +46,24 @@ def heading_text(resource)
...
+    elsif site?
+      Rails.application.routes.url_helpers.deregister_registration_site_path(
+        registration_reference: resource.registration.reference, id: resource.id
+      )
     else
       Rails.application.routes.url_helpers.deregister_exemptions_path(resource.id)
     end
   end
+
+  def successful_redirect_path(resource)
```
jjromeo: It doesn't seem to me that this belongs here as it isn't display logic and is used in the controller only. Maybe it should be in a controller concern, service or controller private method

app/models/waste_exemptions_engine/address.rb
```diff
@@ -11,5 +11,25 @@ class Address
...
     scope :nccc, -> { where("postcode ~* '#{NCCC_POSTCODE_REGEX}'") }
     scope :not_nccc, -> { where.not(id: nccc) }
+
+    def reference
+      site_suffix.present? ? "#{registration.reference}/#{site_suffix}" : registration.reference
+    end
+
+    def site_status
+      if registration.multisite?
+        registration_exemptions.map(&:state).include?("active") ? "active" : "deregistered"
```
jjromeo: Can use `any?` rather than map and then include?

Or could use pluck instead of map?

app/models/waste_exemptions_engine/registration.rb
```diff
@@ -51,6 +51,10 @@ def expired?
       state == "expired"
     end
 
+    def multisite?
```
jjromeo, line 54: Will you remove and rely on the engine version? Or did you want to avoid the feature flag being taken in to account?

spec/models/waste_exemptions_engine/address_spec.rb
```diff
@@ -86,4 +86,136 @@
...
+      it { expect(address.reference).to eq(registration.reference) }
+    end
+
+    shared_examples "returns the registration reference and site suffix" do
+      before { address.update(site_suffix: "ABC") }
+
+      it { expect(address.reference).to eq("#{registration.reference}/ABC") }
+    end
+
+    context "when registration is single-site" do
```
jjromeo, line 105: Hmm is it worth having these specs if multisite status of the registration isn't taken into account in the code?

brujeo: The pages I'm working on use the same controlers/views for single and multi-iste registrations, so these specs test both use cases

app/models/waste_exemptions_engine/registration.rb
brujeo: I think I'll keep it separate in order to avoid taking feature flag into account. The argument for that is that if we have multisite registration created, it's wrong (incorrect) to display it as single site regardless of the feature flag.

app/models/waste_exemptions_engine/address.rb
brujeo: refactored

app/presenters/deregistrations_presenter.rb
brujeo: refactored

app/views/shared/_resource_exemptions_list.html.erb
jjromeo: If you did this with locals[:hide_caption] it would return nil if the variable didn't exist

jjromeo (approved):

app/views/shared/_resource_exemptions_list.html.erb
brujeo: `locals[:hide_caption] ` raises `undefined local variable or method 'locals' for an instance` error

PR #1936, opened 2025-11-13
Feature/ruby 3993 renewal template ids

PaulDoyle-EA:
This refactors renewal reminder services and specs.
https://eaflood.atlassian.net/browse/RUBY-3993


app/models/waste_exemptions_engine/registration.rb
```diff
@@ -57,6 +53,10 @@ def multisite?
       is_multisite_registration == true
     end
 
+    def legacy_bulk_or_multisite?
+      multisite? || is_legacy_bulk
```
brujeo, line 57: we can't use `registration.multisite? ` method in back ofice atm, because it depends on a feature flag. So, if registration is multisite, but the feature flag was switched off for some reason, it would return false and use the wrong template

brujeo: there's a tech debt taks to refactor this weird logic in `registration.multisite?`

app/services/renewal_reminders/first_renewal_reminder_email_service.rb
```diff
@@ -19,7 +19,11 @@ def communications_log_params
     private
 
     def template
-      @registration.is_legacy_bulk? ? "69a8254e-2bd0-4e09-b27a-ad7e8a29d783" : "b1c9cda2-b502-4667-b22c-63e8725f7a27"
+      if @registration.legacy_bulk_or_multisite?
```
brujeo: Alternative approach without nested IF's

```
return "b1c9cda2-b502-4667-b22c-63e8725f7a27" unless @registration.legacy_bulk_or_multisite?

renewals_enabled? ? "cda801d8-ad08-4e77-ab46-94b0e9689ed7" : "69a8254e-2bd0-4e09-b27a-ad7e8a29d783"
```

PR #1956, opened 2025-12-15
Chore/ruby 4088 refactor multisite checks

PaulDoyle-EA:
Align with engine changes to multisite checking.
https://eaflood.atlassian.net/browse/RUBY-4088

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -61,11 +61,18 @@ RSpec/ContextWording:
   Exclude:
     - spec/support/shared_examples/rake.rb
 
-# Simply changing these variables to "let" results in many cases in an error because "let"
-# variables are not available on an example group, only in the scope of an example.
-# TODO: Enable this cop and refactor the relevant 137 instances.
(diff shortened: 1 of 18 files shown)


.rubocop.yml
```diff
@@ -61,6 +61,12 @@ RSpec/ContextWording:
   Exclude:
     - spec/support/shared_examples/rake.rb
 
+# Simply changing these variables to "let" results in many cases in an error because "let"
+# variables are not available on an example group, only in the scope of an example.
+# TODO: Enable this cop and refactor the relevant 137 instances.
+RSpec/LeakyLocalVariable:
+  Enabled: false
+
```
brujeo: I do not think we should disable this rule entirely. I have not encountered any cases where it is genuinely impossible to eliminate leaky local variables. There are scenarios where the rule creates more friction than value, but those should be handled as targeted, case-by-case exceptions rather than a blanket change.

brujeo (approved):

PR #1979, opened 2026-01-14
Move site status labels to i18n and add Expired status

PaulDoyle-EA:
https://eaflood.atlassian.net/browse/RUBY-4133

### DIFF
--- a/app/models/waste_exemptions_engine/address.rb
+++ b/app/models/waste_exemptions_engine/address.rb
@@ -17,7 +17,8 @@ module WasteExemptionsEngine
     end
 
     def site_status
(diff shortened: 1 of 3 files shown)


app/models/waste_exemptions_engine/address.rb
```diff
@@ -17,7 +17,8 @@ def reference
     end
 
     def site_status
-      registration_exemptions.any? { |re| re.state == "active" } ? "active" : "deregistered"
+      # Currently all registration_exemptions have the same status, so just use the latest
+      I18n.t "sites.index.states.#{registration_exemptions.last.state}"
     end
```
brujeo, line 22: I don't think this would work for multisite. All exemptions within a single site have the same status, but other sites might be deregistered

PR #1981, opened 2026-01-16
Fix/ruby 4133 site expired

PaulDoyle-EA:
Use i18n for site address status display and add support for "expired" status.
https://eaflood.atlassian.net/browse/RUBY-4133

### DIFF
--- a/app/models/waste_exemptions_engine/address.rb
+++ b/app/models/waste_exemptions_engine/address.rb
@@ -17,14 +17,13 @@ module WasteExemptionsEngine
     end
 
     def site_status
-      state = if registration_exemptions.any? { |re| re.state == "active" }
-                "active"
-              elsif registration_exemptions.any? { |re| re.state == "expired" }
-                "expired"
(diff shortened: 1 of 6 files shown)


app/models/waste_exemptions_engine/address.rb
```diff
@@ -17,7 +17,14 @@ def reference
...
     def site_status
-      registration_exemptions.any? { |re| re.state == "active" } ? "active" : "deregistered"
+      state = if registration_exemptions.any? { |re| re.state == "active" }
+                "active"
+              elsif registration_exemptions.any? { |re| re.state == "expired" }
+                "expired"
+              else
+                "deregistered"
+              end
+      I18n.t "sites.index.states.#{state}"
```
brujeo: I think a more flexible approach would be to return a “clean” status here and handle the translation in the template.

brujeo (dismissed):

brujeo (approved):

PR #2012, opened 2026-02-18
[RUBY-4194] Fix charge amount calculation for multisite registrations in finance export

brujeo:
WEX - Finance Export - incorrect compliance_initial charge
https://eaflood.atlassian.net/browse/RUBY-4194

- Extracted common logic from InitialComplianceChargeRowPresenter and AdditionalComplianceChargeRowPresenter into a parent class ComplianceChargeRowPresenter
- Fixed charge amount calculation for multisite registrations in finance export
- Fixed an issue with balance not being calculated properly for multisite registrations

### DIFF
--- a/app/presenters/reports/finance_data_report/additional_compliance_charge_row_presenter.rb
+++ b/app/presenters/reports/finance_data_report/additional_compliance_charge_row_presenter.rb
@@ -2,43 +2,17 @@
 
 module Reports
   module FinanceDataReport
-    class AdditionalComplianceChargeRowPresenter < BaseRegistrationRowPresenter
+    class AdditionalComplianceChargeRowPresenter < ComplianceChargeRowPresenter
       def charge_type
         "compliance_additional"
       end
 
-      def charge_amount
-        charge_amount = if @registration.multisite?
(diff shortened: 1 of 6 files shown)


app/presenters/reports/finance_data_report/additional_compliance_charge_row_presenter.rb
```diff
@@ -8,7 +8,14 @@ def charge_type
       end
 
       def charge_amount
-        display_pence_as_pounds_and_pence(pence: @secondary_object.additional_compliance_charge_amount,
+        charge_amount = if @registration.multisite?
```
jjromeo: Is it worth moving this logic into a private method on the class these two presenters inherit form?

brujeo: The base presenter serves as a parent for numerous other presenters that do not require this functionality, so introducing it there would not be appropriate.

We could create a dedicated ComplianceChargeRowPresenter class and have these two presenters inherit from it. However, I feel that would be overengineering at this stage.

jjromeo (dismissed):

jjromeo (approved):

PR #2026, opened 2026-02-27
[RUBY-4201] Add summary and no compliance charge rows to finance data report

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-4201

Alter Finance Data Export to show Current Balance of Registrations

 ## Summary

  - Add a summary row for each order with charge_type = summary
  - `payment_status` is now only populated on the `summary` row (blank on all other rows).
  - `summary.balance` shows the final running balance (current balance).
  - `summary.charge_amount` now shows the total of the charge rows being summarised.
  - Added `compliance_no_charge` rows (`charge_type = compliance_no_charge`, `charge_amount = 0`) when a band has no initial / additional compliance charge.
  - Added back-office comments to the export comments column:
    - payment row comments from `payment.comments`
    - charge adjustment comments from `charge_adjustment.reason`

### DIFF
--- a/app/models/reports/finance_data_report/data_serializer.rb
+++ b/app/models/reports/finance_data_report/data_serializer.rb
@@ -52,6 +52,7 @@ module Reports
         return [] if order&.charge_detail.blank?
 
         @total = 0
+        @charge_total = 0
         charge_detail = order.charge_detail
 
         order_rows = []
@@ -59,6 +60,7 @@ module Reports
         order_rows.concat(generate_compliance_charge_rows(registration, charge_detail))
         order_rows.concat(generate_charge_adjustment_rows(registration))
         order_rows.concat(generate_payment_rows(registration))
+        order_rows.concat(summary_row(registration))
         order_rows
       end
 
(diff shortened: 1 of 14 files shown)


brujeo (commented):
Missing specs for SummaryRowPresenter and ComplianceNoChargeRowPresenter

app/models/reports/finance_data_report/data_serializer.rb
```diff
@@ -167,6 +175,35 @@ def payment_row(registration, secondary_object, site_address)
...
+        end
+        @charge_total += presenter.summary_charge_amount_in_pence
+        @total = presenter.total
+        [output]
+      end
+
+      def summary_row(registration)
+        presenter = FinanceDataReport::SummaryRowPresenter.new(
+          registration:,
+          secondary_object: @charge_total,
```
brujeo: Not a big issue, just a bit smelly.... overloading an object with an integer value.

app/models/reports/finance_data_report/data_serializer.rb
```diff
@@ -167,6 +175,35 @@ def payment_row(registration, secondary_object, site_address)
...
+          total: @total,
+          show_payment_status: true
```
brujeo: The summary row is structurally different enough that it probably shouldn’t inherit from BaseRegistrationRowPresenter.

Separating it would let us avoid workarounds such as show_payment_status: true and similar flags that only exist to bend the base presenter’s behaviour.

brujeo (approved):

PR #2059, opened 2026-04-09
[RUBY-4189] Display free exemptions on their own row in payment details

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-4189

- Display free exemptions in their own row in payment details
- Refactor code surrounding payment details and charge details to use a presenter

image


app/presenters/payment_details/charge_breakdown_presenter.rb
```diff
@@ -0,0 +1,107 @@
...
+
+      def amount_cell_classes
+        return "govuk-table__cell govuk-!-text-align-right govuk-!-padding-top-0" if @total
+        return "govuk-!-text-align-right govuk-!-padding-top-2" if @top_padded
+
+        "govuk-!-text-align-right"
+      end
+    end
+
+    def initialize(order:, multisite:)
```
brujeo: it's not clear that multisite is boolean. Also, if multisite is boolean, why not to have it set to false by default? 
`def initialize(order:, is_multisite: false)`

brujeo (approved):

PR #2082, opened 2026-05-11
[RUBY-4263] Back office edit workflow moved from engine to back office app

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-4263

Moves the back-office edit registration flow into the back-office app, so the BO app owns the BO-only routes, controllers, forms, views, workflow and specs.

This keeps the shared engine for shared behaviour, but removes the BO edit entry points from the engine so they are no longer present in the front-office application by default.

The existing transient registration class name is still WasteExemptionsEngine::BackOfficeEditRegistration, and the persisted back_office_* workflow states are also unchanged, to avoid introducing a data migration at this stage.

The BO app now uses route/controller/form names without "Back Office", so there is a mapping to map the existing workflow states to the new BO route/controller names.

Because some shared forms used in the back office edit registration process still live in the engine, the BO app prepends CanRedirectEditRegistration so BackOfficeEditRegistration submissions redirect to BO app routes rather than engine routes.


app/forms/waste_exemptions_engine/registration_number_form.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    private
+
+    def update_edit_registration_details(params)
+      company_no = process_company_no(params[:temp_company_no])
+      companies_house_details = DefraRuby::CompaniesHouse::API.run(company_number: company_no)
+
+      params[:operator_name] = companies_house_details[:company_name]
+      params[:company_no] = company_no
+    end
```
brujeo, line 30: In the original engine code, company_no was zero-padded regardless of the transient registration type. Probably makes sense to keep the same behaviour

Image

app/controllers/edit_forms_base_controller.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+class EditFormsBaseController < WasteExemptionsEngine::FormsController
+  include EditPermissionChecks
+  prepend CanRedirectEditRegistration
```
brujeo: No explicit prepending for CanRedirectEditRegistration is needed as it gets already prepended for every single form in config/initializers/waste_exemptions_engine.rb

brujeo (approved):

PR #2089, opened 2026-05-27
RUBY 4299 wex report for sscl receipts

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-4299

- Add SSCL receipts report generation and export functionality
- Add SSCL reports to scheduler


app/services/reports/sscl_receipts_report/export_service.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require_relative "../../concerns/can_load_file_to_aws"
```
brujeo: I think autoloader should load it automatically

brujeo (approved):

PR #2126, opened 2026-07-21
Feature/ruby 4364 wex bo identify and fix broken registration data

brujeo:
https://eaflood.atlassian.net/browse/RUBY-4364

Added 2 rake tasks to cleanup broken registration data:

1. cleanup:identify_registrations_without_exemptions_or_sites - Identifies registrations that are missing exemptions and/or site addresses

2. cleanup:delete_registration - Deletes a single registration (and all of its linked records) identified by its reference, but ONLY when it has no registration exemptions

### DIFF
--- a/app/lib/command_line_table_formatter.rb
+++ b/app/lib/command_line_table_formatter.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+# Formats a collection of row hashes as a fixed-width, aligned plain-text table
+# for command-line / rake output. The keys of the first row become the column
+# headers (in order); each row is read by those keys and its values stringified.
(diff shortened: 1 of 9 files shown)


app/lib/command_line_table_formatter.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+# Formats a collection of row hashes as a fixed-width, aligned plain-text table
```
jjromeo, line 3: Cool!

jjromeo (dismissed):
Looks good!

jjromeo (approved):

